PR #6, opened 2021-07-18
logd refactor

brandonbloom:
This is a pretty big refactor of the logd package and associated command line tools.

The _original_ goal here was to introduce some new mechanism for handling when logd isn't running during testing, or if some failure occurs and the add/remove-logs get out of sync. That is, we need some kind of reconciliation loop. However, I didn't actually achieve that in this PR. As I was working on it, I ran in to too many design problems with the existing concurrency, so I completely re-worked that. In the interest of not making this impossible to review, I stopped there before addressing the sync/convergence issues.


cmd/logd/main.go
```diff
@@ -14,36 +14,39 @@ import (
...
-	}
 
-	go func() {
-		c := make(chan os.Signal, 1)
-		signal.Notify(c, os.Interrupt)
-		<-c
-		collector.Stop(ctx)
-		os.Exit(0)
-	}()
+	{
```
brandonbloom, line 35: My first attempt at this was to do an error group or something like that to run the log collector and http server in parallel. In that world, they share a cancellation context, but that's not right! You actually want them nested, so that on interrupt, the http server shutsdown _first_, and then the collector shuts down. If you do it in parallel, you have a race condition where some http requests may attempt to access a collector that has already been shutdown.

util/cmdutil/http.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+func ShutdownOnInterrupt(ctx context.Context, svr *http.Server) {
+	ctx, stop := signal.NotifyContext(ctx, os.Interrupt)
+	defer stop()
+
+	<-ctx.Done()
+	stop()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
brandonbloom, line 29: 5 seconds ought to be enough for anybody :)

util/agent/agent.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"context"
+	"errors"
+)
+
+// Agent guards asynchronous changes of an individual location.
+// Inspired by <https://clojure.org/reference/agents>.
+// Note: This is not quite a Clojure agent, since there is no way to deref.
+type Agent struct {
```
brandonbloom, line 11: I've written something like this about ~20 times in Go-world, and it's always subtle and difficult to get right. This is my only ever successful attempt generalizing it to something reusable. Not 100% satisfied with it, but at least it seems correct in isolation.

brandonbloom:
Going to merge, can do post-hoc review if you like.

gensym/ulid.go
```diff
@@ -1,30 +1,30 @@
...
 	randRead := rand.New(rand.NewSource(seed))
-	return &idGen{
+	return &ULIDGenerator{
 		entropy: ulid.Monotonic(randRead, 0),
 	}
 }
 
-func (gen *idGen) nextId(ctx context.Context) ([]byte, error) {
+func (gen *ULIDGenerator) NextID(ctx context.Context) ([]byte, error) {
 	gen.mu.Lock()
```
kendru, line 28: Oh, btw, this lock is in here because the math/rand source is not thread-safe :(

logd/server/collector.go
```diff
@@ -0,0 +1,308 @@
...
+func (lc *LogCollector) stopWorker(logName string) {
+	wkr := lc.workers[logName]
+	if wkr == nil {
+		return
+	}
+	wkr.stop()
+}
+
+func (lc *LogCollector) DescribeLogs(ctx context.Context, input *api.DescribeLogsInput) (output *api.DescribeLogsOutput, err error) {
+	err = lc.agent.Send(func() error {
```
kendru, line 225: Okay, this pattern is kinda nice :)

kendru: It would be nice to be able to use capabilities of some sort to guard access to describeLogs outside of agent ;)

util/atom/atom.go
```diff
@@ -1,6 +1,7 @@
 package atom
 
-// An uncoordinated, atomic reference cell in the style of Clojure.
+// Atom are uncoordinated, atomic reference cells.
+// Inspired by <https://clojure.org/reference/atoms>.
```
kendru, line 4: So are you just rewriting Clojure now? haha

util/cmdutil/http.go
kendru: riiiiiight :D

kendru (commented):
Great change!

logd/server/collector.go
brandonbloom: It took me a long while to get to this pattern! I think this is the least-bad, memory-model/concurrency-safe approach that doesn't require reflection.

brandonbloom: RE: Capabilities. Yeah, absolutely. I have some ideas on that front. It's starting to become relevant as I get moving on multiple-workspaces too. Will discuss in that PR when it comes up.

util/atom/atom.go
brandonbloom: I mean, Clojure got sooo so much stuff right.

PR #9, opened 2021-07-19
Workspaces

brandonbloom:
OK, this now works and has `exo gui` integration and there is a CreateWorkspacePage.

Code could use a solid cleanup pass, but probably ready to land for now.


exod/main.go
```diff
@@ -18,6 +19,10 @@ func Main() {
 		MuxPattern: "/",
 	}
 
+	if err := os.Chdir("/"); err != nil {
```
kendru, line 24: It would probably be good to comment this.

PR #10, opened 2021-07-19
Add Telemetry notice to README

jwmza:
Not sure if this is the right place but pretty sure, please confirm:

@kendru @brandonbloom

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,3 +28,14 @@ For more features, consult the builtin help:
 ```bash
 exo help
 ```
+
+---
+
+## Telemetry
+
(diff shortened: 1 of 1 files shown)


kendru:
Works for me

README.md
```diff
@@ -28,3 +28,14 @@ For more features, consult the builtin help:
...
 ```
+
+---
+
+## Telemetry
+
+**exo** collects limited and anonymous telemetry data by default. This behavior can be disabled by adding the following setting to your exo config (located at `~/.exo/config.toml` by default):
+```bash
+[telemetry]
+enabled: false
```
brandonbloom, line 39: Do we want to make this `disable: true` so that true/false/nil logic works out with the appropriate default?

kendru: Yeah, I think that's the better option.

Issue #4, opened 2021-07-16
Self-install release process


kendru:
Implement the process from the installtest repo for delivering a binary release to users.

PR #11, opened 2021-07-19
Add self-update command

kendru:


telemetry/autoinstalled.go
```diff
@@ -0,0 +1,45 @@
...
+
+	if err := os.Chmod(tmpfile.Name(), 0700); err != nil {
+		return fmt.Errorf("making script executable: %w", err)
+	}
+
+	out, err := exec.Command(tmpfile.Name()).Output()
+	if err != nil {
+		return fmt.Errorf("running update script: %w", err)
+	}
+	fmt.Printf(string(out))
```
brandonbloom, line 42:
```go
cmd.Stdout = os.Stdout
cmd.Stderr = os.Stderr
```

core/version.go
```diff
@@ -2,5 +2,10 @@ package core
 
 import _ "embed"
 
+const (
+	CheckVersionEndpoint = "https://download-page.deref.workers.dev/version.txt"
+	UpdateScriptEndpoint = "https://download-page.deref.workers.dev/install.sh"
```
brandonbloom, line 7: We don't control `workers.dev`, so let's use the redirecting endpoint to insulate ourselves from the implementation service.

brandonbloom (approved):
LGTM except we should use URLs for which we fully control the domain.

kendru:
Fixes #4

PR #20, opened 2021-07-21
minimal error message handling

brandonbloom:
This isn't ideal, as random internal bits of code has to have knowledge about public vs private errors, but it's a reasonable heuristic. Basically, any error that isn't explicitly assigned an HTTP status code is treated as a 500 and the error message is logged on the server, but not returned to the client.

Example of the UI looks like this:

image


brandonbloom:
merging, but post-hoc review of course welcome :)

components/process/process.go
```diff
@@ -61,7 +63,7 @@ func (provider *Provider) start(ctx context.Context, procDir string, inputSpec s
 	}
 	program, err := whichQ.Run()
 	if err != nil {
-		return state{}, err
+		return state{}, errutil.WithHTTPStatus(http.StatusBadRequest, err)
```
kendru, line 66: I agree that having internals know about http is not ideal. One pattern that I like is having a set of domain errors constructors that internally set an http status code, e.g. `NewValidationFailedErr()` would create an error object that had an `HTTPStatusCode: 400`.

kendru (commented):
Looks good to me. I think that this is a reasonable approach. In general, I think that we should try to make more pervasive use of the `RequestData` type (and potentially fix problems with it), but that might make more sense after we codegen an API client.

components/process/process.go
brandonbloom: yup. right now, this is very much "do something that works and make sure all the badness has an easy-to-grep for string" :)

brandonbloom:
I'm not familiar. What do you mean by "RequestData"?

PR #23, opened 2021-07-21
syslog

brandonbloom:
switch to syslog instead of log collector workers reading fifos


brandonbloom:
@kendru please review.

@BenElgar if you have log handling issues on linux, please try this branch

logd/server/collector.go
```diff
@@ -5,232 +5,92 @@ import (
...
-	return jsonutil.MarshalFile(lc.statePath, &lc.state)
-}
-
-func (lc *LogCollector) Run(ctx context.Context) error {
-	lc.workers = make(map[string]*collectorWorker)
-
-	if err := lc.loadState(); err != nil {
-		return fmt.Errorf("loading state: %w", err)
+func (lc *LogCollector) AddEvent(ctx context.Context, input *api.AddEventInput) (*api.AddEventOutput, error) {
+	if input.Log == "" {
```
kendru, line 39: Do we handle pointer-type inputs? Not that there is a case for a log that is an empty string, but I'm not clear on the semantics of `*string`  vs `string` input types.

logd/store/badger/history.go
```diff
@@ -164,6 +165,9 @@ func getEventWithCursors(event api.Event) (store.EventWithCursors, error) {
 	return eventWithCursors, nil
 }
 
+// TODO: RemoveOldEvents also needs to delete events over a certain age,
+// not just a count limit. This is necessary for cleanup of no-longer-tracked
+// log streams.
 const maxEventsPerStream = 5000
```
kendru, line 171: I am considering indexing the logs on `(id, log)` instead of `(log, id)` since I think that it will be more common to try to scan (most or all) events in a given timeframe than to scan events from one specific log. That would allow the "delete older than" operation to be more efficient.
Either way, we'll need a secondary index, so I'll add that soon.

logd/store/api.go
```diff
@@ -2,17 +2,21 @@ package store
 
 import (
 	"context"
-	"io"
 
 	"github.com/deref/exo/logd/api"
 )
 
 type Store interface {
+	// Returns the next log after the given argument.
```
kendru, line 10: It would be helpful to define what "next" means in this context. My immediate question is whether this would iterate through all logs available (regardless of what log I passed in first), or would or follow some natural ordering, potentially terminating without iterating all logs if I did not start at the "beginning".

logd/store/badger/store.go
```diff
@@ -45,3 +45,42 @@ func (sto *Store) GetLog(name string) store.Log {
...
+
+func (sto *Store) NextLog(after store.Log) (store.Log, error) {
+	prefix := []byte{}
+	if after != nil {
+		prefix = append([]byte(after.Name()), 255)
+	}
+	var nextName string
+	if err := sto.db.View(func(txn *badger.Txn) error {
+		opts := badger.DefaultIteratorOptions
+		opts.PrefetchSize = 1000
```
kendru, line 57: This option is not used with `PrefetchValues = false`

logd/store/badger/store.go
```diff
@@ -45,3 +45,42 @@ func (sto *Store) GetLog(name string) store.Log {
...
+		it := txn.NewIterator(opts)
+		defer it.Close()
+
+		// TODO: Something better than a linear key scan.
+		it.Seek(prefix)
+		for it.Valid() {
+			it.Next()
+			if !it.ValidForPrefix(prefix) {
+				if it.Valid() {
+					nextName = string(it.Item().Key())
```
kendru, line 69: Is this correct? If I'm reading this right, we are looking for a log with the same name as the key, but the key is `(log + <255>, id).

logd/store/badger/store.go
```diff
@@ -45,3 +45,42 @@ func (sto *Store) GetLog(name string) store.Log {
...
+	var nextName string
+	if err := sto.db.View(func(txn *badger.Txn) error {
+		opts := badger.DefaultIteratorOptions
+		opts.PrefetchSize = 1000
+		opts.PrefetchValues = false
+
+		it := txn.NewIterator(opts)
+		defer it.Close()
+
+		// TODO: Something better than a linear key scan.
```
kendru, line 63: Did somebody say secondary indexes? ;)

logio/main.go
```diff
@@ -0,0 +1,169 @@
+package logio
```
kendru, line 1: Instead of naming this `logio`, I think that we should go ahead and name it something more generic, since this is the component that will supervise processes. Maybe we should use something that conveys the idea of "supervisor", "proxy" or "wrapper" - foreman? bigbrother? cocoon? chrysalis? saran 😆?

kendru (approved):

chrono/chrono.go
```diff
@@ -6,7 +6,8 @@ import (
 )
 
 const (
-	RFC3339NanoUTC = "2006-01-02T15:04:05.999999999Z"
+	RFC3339NanoUTC  = "2006-01-02T15:04:05.999999999Z"
+	RFC3339MicroUTC = "2006-01-02T15:04:05.999999Z"
```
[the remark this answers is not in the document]
brandonbloom: Probably?

cmd/exop/main.go
```diff
@@ -39,8 +39,9 @@ func main() {
 	store := statefile.New(statePath)
 
 	cfg := &server.Config{
-		VarDir: paths.VarDir,
-		Store:  store,
+		VarDir:     paths.VarDir,
+		Store:      store,
+		SyslogAddr: "localhost:4500", // XXX Configurable?
```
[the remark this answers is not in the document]
brandonbloom: Yeah, I didn't check if this port was available. I'm not quite sure what the right thing to do here is. I actually think maybe we should pick a port at random and write it in the run state / config file thing. What do you think?

logd/README.md
```diff
@@ -1,44 +1 @@
...
-Use the `logd` command.
-
-## Log Rotation
-
-*NOT YET IMPLEMENTED*
-
-# Development
-
-See `logp` command.
+Provides a UDP-based Syslog server.
```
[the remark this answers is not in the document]
brandonbloom: Heh, I meant to delete this file as it was all out of date and I didn't really have much interesting to write that wouldn't have been better on the interface or similar.

logd/logd.go
```diff
@@ -0,0 +1,116 @@
+package logd
```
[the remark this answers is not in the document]
brandonbloom: Yeah, so much better. I absolutely love the "all public fields and otherwise stateless" struct pattern.

logd/logd.go
```diff
@@ -0,0 +1,116 @@
...
+	if rfc5425Message.Appname == nil {
+		return nil, errors.New("expected APP-NAME")
+	}
+	if rfc5425Message.MsgID == nil {
+		return nil, errors.New("expected MSGID")
+	}
+	if rfc5425Message.Timestamp == nil {
+		return nil, errors.New("expected TIMESTAMP")
+	}
+
```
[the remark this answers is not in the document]
brandonbloom: ¯\_(ツ)_/¯ at some point, I guess :-P

logd/server/collector.go
brandonbloom: No hard and fast rule yet. In this case, empty and nil are both invalid, so this handles both via a combo of static and dynamic means. Need more general validation logic though as we get stricter about this, I imagine.

logd/store/api.go
brandonbloom: Good point. I'll improve this comment.

logd/store/badger/history.go
brandonbloom: 👍🏻

logd/store/badger/store.go
brandonbloom: D'oh. OK.

logd/store/badger/store.go
brandonbloom: D'oh, you're right, this makes no sense at all. Will fix.

logio/main.go
brandonbloom: Yeah, definitely needs a better name. I was thinking about just calling it `supervise`. That work?

cmd/exop/main.go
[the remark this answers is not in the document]
kendru: I think that is the right default behaviour, but it would still be nice to let users specify a specific port if they want to for whatever reason. This is another place where I think that it would be useful for `exo` to own the definition for the `LogProvider`. That way, there is one source of truth for the proxy to discover where to send logs.

chrono/chrono.go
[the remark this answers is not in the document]
kendru: Definitely not a big deal now, but for our purposes, I'm guessing that we don't need to record to nanosecond precision 🤷

PR #26, opened 2021-07-22
Add colored processes and logs with hashed IDs

jwmza:
Early working version:

image

Also adds active state to buttons:

image

Also adds a **GitHub feedback button** to the top right of the header, and adds tooltips to icon buttons.


gui/src/components/IconButton.svelte
```diff
@@ -1,4 +1,11 @@
-<button on:click>
+<script lang="ts">
+
+export let activeState: boolean = false;
```
brandonbloom, line 3: nit: can simply call this "active"

jwmza:
Coloring still has a long way to go but this unit of changes is good enough to be wrapped up here - going to merge & loop around in the next handful of days to synchronize process colors around the app and improve conflict issues.

PR #32, opened 2021-07-26
Add back button to creation form page headers

jwmza:
This adds navigation functionality to the layout header.

image

The **Exo logo** takes you to the home page `/`, and the optional back button by default also takes you home, or to a custom `backButtonRoute`.


gui/src/components/Layout.svelte
```diff
@@ -4,16 +4,32 @@ import IconButton from './IconButton.svelte';
 import * as router from 'svelte-spa-router';
 
 import Feedback from './mono/feedback.svelte';
+import GoBack from './mono/leftarrow.svelte';
+
+export let showBackButton: boolean = false;
```
brandonbloom, line 9: What's the purpose of having separate props for both showBackButton and backButtonRoute? Seems like you could make the default value of backButtonRoute undefined or empty string, and then you only need to specify that instead of both the route and the boolean flag.

PR #35, opened 2021-07-26
gui list of workspaces on home page

brandonbloom:
WIP for @jwmza to takeover and finish

### DIFF
--- a/gui/src/lib/api.ts
+++ b/gui/src/lib/api.ts
@@ -145,11 +145,20 @@ export interface ProcessSpec {
   environment?: Record<string, string>;
 }
 
+export interface WorkspaceDescription {
+  id: string;
+  root: string;
+}
+
 export const api = (() => {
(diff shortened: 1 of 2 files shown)


jwmza:
Auto-responsive grid layout to prevent ridiculously long boxes on big screens and allow devs to have dozens of workspaces if they so choose in a layout that also looks good with just a few!

image

brandonbloom:
What is `default` here? Seems like you might have some leftover gunk from before workspaces were properly implemented. I suggest you use run `exo exit` and then `rm -rf ~/.exo`.

Also, my expectation is that most users will have one to a single-digit number of workspaces, but I guess in theory some users may have quite a few.

brandonbloom:
i'm going to merge this as it's better than not having it, but i think the design needs another pass. thanks @jwmza

jwmza:
Agreed thanks @brandonbloom

PR #36, opened 2021-07-26
`exo run`

brandonbloom:
This is _mostly_ right. It _always_ stops the workspace after a run. If the workspace is already running, it sort of "takes over" in the sense that it doesn't care if anyone else is running already. It will immediately apply the given manifest, then tail logs, and then stop the workspace on interrupt.

Future work would handle some sort of mutual exclusivity, refusing to run if already started. However, I think this works well enough for initial release.

Note: it's easier to review this if you set the diff tool to ignore whitespace changes.

### DIFF
--- a/cmd/exo/apply.go
+++ b/cmd/exo/apply.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"io/ioutil"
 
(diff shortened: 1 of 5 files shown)


brandonbloom:
draft until #50 lands

brandonbloom:
ok, works now, updated to merge against same branch as #50 ready for review @BenElgar or @kendru

PR #41, opened 2021-07-27
Add check for nil message

BenElgar:
This was causing nil pointer exceptions when trying to start this simple heroku example: https://github.com/heroku/node-js-getting-started


brandonbloom:
No brainer, merged, thanks.

How does this actually happen though? logging a blank line?

PR #43, opened 2021-07-27
Change SIGKILL to SIGINT

BenElgar:


brandonbloom:
We probably shouldn't do this unless we also have a force kill after a timeout.

PR #48, opened 2021-07-27
Core vs exod

brandonbloom:
Some small refactors so that that simplify some other refactors coming in order to finish up #36


kendru (approved):
LGTM!

PR #49, opened 2021-07-27
IDL model

brandonbloom:
This PR does two major things:

(A) This splits the existing IDL codegen bits in to three parts: 1) parsing the HCL 2) a deeply interlinked append-only "model" and 3) the actual code templating
(B) Introduces a notion of "extends" a la Go interface embedding.

The model supports interface embedding.

Motivation: Allows workspace to extend process (in a follow-up PR), so that I can start/stop/restart a whole workspace.

### DIFF
--- a/cmd/codegen/main.go
+++ b/cmd/codegen/main.go
@@ -9,6 +9,8 @@ import (
 	"strings"
 
 	"github.com/deref/exo/josh/codegen"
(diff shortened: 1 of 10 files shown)


brandonbloom:
OK, actually, will push one more commit to finish the extends work.

PR #51, opened 2021-07-28
Add upgrade to UI

kendru:
Adds the current version to a footer in the UI. If a new version is available, the user can click a button to request the server perform a self update.

The UI and UX here can be improved, but the basics are working.

When the server performs an update, it downloads a new binary then attempts to replace the current process with the newly-downloaded version. We need to take an extra step to resolve the _new_ binary, since the actual binary that is running is likely a versioned executable, and the new executable is also versioned with a different name.


brandonbloom:
Sorry about the conflicts! They should be trivial moves/renames, so easy to resolve.

exod/server/kernel.go
```diff
@@ -77,10 +82,56 @@ func (kern *Kernel) GetVersion(ctx context.Context, input *api.GetVersionInput)
...
+	if isSymlink, _ := osutil.IsSymlink(symlinkPath); isSymlink {
+		dest, err := os.Readlink(symlinkPath)
+		if err != nil {
+			exitWithError(err)
+		}
+		dest = path.Join(dir, dest)
+		cmd = path.Clean(dest)
+	}
+
+	if err := syscall.Exec(cmd, append([]string{cmd}, os.Args[1:]...), os.Environ()); err != nil {
```
brandonbloom, line 134: I think we should probably do something like:

```go
if os.Args[1] != "server" {
  exitWithError(errors.New("attempt to restart a non-server exo process"))
  return
}
```

exod/server/kernel.go
```diff
@@ -77,10 +82,56 @@ func (kern *Kernel) GetVersion(ctx context.Context, input *api.GetVersionInput)
...
 	message := input.Message
 	if input.Message == "" {
 		message = "test error"
 	}
 	panic(message)
 }
+
+// restart replaces the current process with a new copy of itself. This is useful
+// after downloading a new binary.
+func restart(ctx context.Context) {
```
brandonbloom, line 106: Generally, it's a tad weird that this is in the "server" package, where most/all of the other CLI functionality is confined to the `cmd` directory of the `exod` package. However, I can't really think of a better way to do this, so I think the layering violation is justified. You might want to do a NOTE or something to cross reference the "server" command, the "upgrade" command, and this code.

gui/src/components/Layout.svelte
```diff
@@ -42,6 +43,9 @@
   <div>
     <slot />
   </div>
+  <footer>
```
brandonbloom, line 46: Do we want this in the footer all the time?

We should probably stuff it in to a menu somewhere once we have a menu somewhere to stuff it :)

gui/src/components/VersionInfo.svelte
```diff
@@ -0,0 +1,41 @@
...
+  import { onDestroy, onMount } from 'svelte';
+  import { api } from '../lib/api';
+  import Button from './Button.svelte';
+
+  let installedVersion: string | null = null;
+  let latestVersion: string | null = null;
+
+  const doUpgrade = async () => {
+    await api.kernel.upgrade();
+  }
```
brandonbloom, line 11: missing semicolon. Did prettier run?

gui/src/components/VersionInfo.svelte
```diff
@@ -0,0 +1,41 @@
...
+    if (fetchTimeout !== null) {
+      clearTimeout(fetchTimeout);
+    }
+  });
+</script>
+
+<section>
+  exo {installedVersion || ''}
+  {#if latestVersion !== null}
+    | Update available: <strong>{latestVersion}</strong> <Button on:click={doUpgrade}>Get It!</Button>
```
brandonbloom, line 39: How long does upgrading take? Surely it's not instantaneous. Should we hide this button or otherwise show status if presently upgrading?

brandonbloom (approved):
LGTM! All comments minor.

Issue #47, opened 2021-07-27
pass config to supervise process on stdin


brandonbloom:
between https://github.com/deref/exo/pull/42/files and https://github.com/deref/exo/issues/44 we should just give up on command line flags and pass supervisor config as json on stdin

brandonbloom:
@BenElgar this is related to #44, so giving it to you. Here's how far I got :-P

```go
type Config struct {
	ComponentID      string   `json:"componentId"`
	SyslogAddress    string   `json:"syslogAddress"`
	WorkingDirectory string   `json:"workingDirectory"`
	Program          string   `json:"program"`
	Arguments        []string `json:"arguments"`
}
```

I think the only other one would be like like `ShutdownTimeout int` in millis or something.

PR #53, opened 2021-07-28
Sigkill after grace period

BenElgar:


cmd/tick/main.go
```diff
@@ -4,10 +4,24 @@ package main
...
+	"os/signal"
+	"syscall"
 	"time"
 )
 
 func main() {
+	c := make(chan os.Signal)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
+	go func() {
+		for sig := range c {
```
brandonbloom, line 17: I believe that will prevent the `tick` program from terminate on ^c in the terminal. This little command line tool is too useful to make it broken :)

If you were just doing this for testing one-off, I'd revert this change. If it's useful going forward, we should make another tool like "hang" or similar.

components/process/provider.go
```diff
@@ -6,14 +6,16 @@ type Provider struct {
...
 type Spec struct {
-	Directory   string            `json:"directory"`
-	Program     string            `json:"program"`
-	Arguments   []string          `json:"arguments"`
-	Environment map[string]string `json:"environment"`
+	Directory                  string            `json:"directory"`
+	Program                    string            `json:"program"`
+	Arguments                  []string          `json:"arguments"`
+	Environment                map[string]string `json:"environment"`
+	ShutdownGracePeriodSeconds int               `json:"shutdownGracePeriodSeconds"`
```
brandonbloom, line 13: This is a mouthful. Why not call it `shutdownTimeout`?

Also: Why seconds and not milliseconds? Seems like 200 ms would be more than enough grace period for many use cases.

supervise/main.go
```diff
@@ -78,31 +82,35 @@ MSGID = The message "type". Set to "out" or "err" to specify which stdio
...
+				if err := cmd.Process.Signal(syscall.SIGTERM); err != nil {
 					break
 				}
+				// After some timeout, check if the process is still running and, if so,
+				// send a SIGKILL.
+				time.Sleep(time.Second * time.Duration(timeout))
+				cmd.Process.Kill()
 			// Exit when child exits.
 			case syscall.SIGCHLD:
 				os.Exit(1)
```
brandonbloom, line 109: is this now potentially the wrong status code for the supervisor?

supervise/main.go
```diff
@@ -78,31 +82,35 @@ MSGID = The message "type". Set to "out" or "err" to specify which stdio
 		fatalf("reporting pid: %v", err)
 	}
 
+	// Dial syslog.
+	conn, err := net.DialUDP("udp", nil, udpAddr)
```
brandonbloom, line 86: curious why this moved before the signal handling. I _think_ DialUDP will never block, but my intuition is to always configure signal handlers as early as possible in case blocking does happen.

supervise/main.go
```diff
@@ -24,7 +24,7 @@ var varDir string
 
 func Main(command string, args []string) {
 	if len(args) < 4 {
-		fatalf(`usage: %s <address> <component-id> <working-directory> <program> <args...>
+		fatalf(`usage: %s <address> <component-id> <working-directory> <timeout> <program> <args...>
```
brandonbloom, line 27: Yeah, we should definitely do #47

supervise/main.go
```diff
@@ -45,8 +45,12 @@ MSGID = The message "type". Set to "out" or "err" to specify which stdio
 	address := args[0]
 	componentID := args[1]
 	wd := args[2]
-	program := args[3]
-	arguments := args[4:]
+	timeout, timeoutErr := strconv.Atoi(args[3])
+	if timeoutErr != nil {
+		panic(timeoutErr)
```
brandonbloom, line 50: Use fatalf so that the error can be more easily reported back to the code that started the supervisor.

components/process/process.go
```diff
@@ -41,7 +41,7 @@ func (provider *Provider) Start(ctx context.Context, input *core.StartInput) (*c
 }
 
 func (provider *Provider) start(ctx context.Context, componentID string, inputSpec string) (State, error) {
-	var spec Spec
+	spec := Spec{ShutdownGracePeriodSeconds: 5}
```
brandonbloom, line 44: This seems really subtle! I was confused at first how this was even working (assuming it is).

Seems like this will default the field when `UnmarshalString` runs only by virtue of the fact that if the field is missing, then it won't get overridden! It's not obvious to me that it wouldn't potentially be overwritten to 0.

A much clearer thing to do would be to change the field to a pointer type and then explicitly handle defaulting when the timeout is nil.

supervise/main.go
```diff
@@ -78,31 +82,35 @@ MSGID = The message "type". Set to "out" or "err" to specify which stdio
...
 		for sig := range c {
 			switch sig {
 			// Forward signals to child.
-			case os.Interrupt, os.Kill:
-				if err := cmd.Process.Signal(sig); err != nil {
+			case os.Interrupt, syscall.SIGTERM:
+				if err := cmd.Process.Signal(syscall.SIGTERM); err != nil {
 					break
 				}
+				// After some timeout, check if the process is still running and, if so,
```
brandonbloom, line 103: This comment suggests some checks/conditions, but it seems to um.. be unconditionally :-P Tho I think unconditional is fine here, assuming no immediate pid reuse (and I think there is nothing better you can do beyond that).

components/process/process.go
```diff
@@ -72,6 +72,7 @@ func (provider *Provider) start(ctx context.Context, componentID string, inputSp
 			provider.SyslogAddr,
 			componentID,
 			provider.WorkspaceDir,
+			fmt.Sprintf("%d", spec.ShutdownGracePeriodSeconds),
```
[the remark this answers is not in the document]
BenElgar: No this is a conversion to a string

cmd/tick/main.go
BenElgar: Fair enough

components/process/provider.go
BenElgar: I basically took the name from Kubernetes which calls it the same thing [0]. `shutdownTimeout` sounds a bit like the maximum amount of time that the process should be allowed to run for.

There aren't any use cases which I can think of where 200ms would be better than 0 or 1 second. If a process intercepts the signal it is presumably going to do some I/O which I would always have more than a 1 second timeout for. For context, the default Kubernetes grace period is 30 seconds.

[0] https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#hook-handler-execution

supervise/main.go
BenElgar: Fixed — I removed the check when I was debugging and forgot to add it back.

components/process/process.go
BenElgar: Discussed in person but it seems reasonably common. It's first answer in both of the top two results on Google for "go json unmarshal default".
https://eli.thegreenplace.net/2020/optional-json-fields-in-go/#setting-default-values
https://stackoverflow.com/a/30445480

supervise/main.go
BenElgar: :+1:

supervise/main.go
BenElgar: Ah this is a leftover from when I was using the connection for printing debug messages. I moved the signal handling above even starting the process. The case in which the signal occurs before the process has been started seems to be less important than the case when the process has been started but signal forwarding hasn't been set up.

components/process/process.go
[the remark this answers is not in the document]
brandonbloom: haha whoops misread :)

supervise/main.go
BenElgar: Fixed

PR #59, opened 2021-07-29
Enable strict mode

BenElgar:


gui/src/lib/process/store.ts
```diff
@@ -35,26 +36,29 @@ export const fetchProcesses = (workspace) => {
       case 'success':
         refetchProcesses(workspace);
         return refetchingResponse(req.data);
+      case 'refetching':
+        // I'm not sure this is right.
+        return req;
```
BenElgar, line 41: @kendru is this the correct thing to do here?

gui/src/lib/api.ts
```diff
@@ -151,6 +151,30 @@ export interface WorkspaceDescription {
   root: string;
 }
 
+export interface WorkspaceApi {
```
brandonbloom, line 154: we should extend our code generator to produce these too :)

gui/src/lib/process/store.ts
brandonbloom: seems reasonable to me, but i think this whole remote data / fetching mechanism needs some love

brandonbloom (approved):
Thanks for doing this!

PR #60, opened 2021-07-29
Process list table with checkboxes

jwmza:
This is an initial visual / DOM overhaul of the ProcessList component to move towards an improved UX with:

**This PR**
- Column-labeled checkboxes for more intuitive interaction with toggles (e.g. log visibility toggler)
- Table layout
- Right-hand process run/stop controls and status

**Upcoming**
- Improved contextual display of controls and information
    - Hide delete button and future similar controls behind hover and/or a dropdown menu
    - Hide play/pause controls behind hover, showing only process status by default in their place
        - Spotify/Apple Music song list -esque UI
    - Show log color in process status display

image


jwmza:
I suggest we merge this as-is to quickly improve the worst part of the former UX which is the log toggle controls. Will follow up with dropdown menus and the nice colored status/action-on-hover controls for process running and stopping.

gui/src/components/CheckboxButton.svelte
```diff
@@ -0,0 +1,81 @@
+<script lang="ts">
+  export let active: boolean = false;
+  export let tooltip: string = '';
+
+  import Show from './mono/check.svelte';
```
brandonbloom, line 5: nit: imports should go at the top of the script block. I'm actually surprised this compiles at all, since this means that the import side-effects are potentially not statically analyzable.

brandonbloom (approved):
LGTM

gui/src/components/CheckboxButton.svelte
jwmza: Thanks for catching this!

PR #62, opened 2021-07-29
Load global config

kendru:
This is a prerequisite to honoring the telemetry disable flag.


brandonbloom:
D'oh! I though we were honoring that already. Doesn't feel great to be misleading about that. Let's prioritize getting this fixed ASAP.

cmd/exo/root.go
```diff
@@ -20,12 +27,20 @@ For more information, see https://exo.deref.io`,
 }
 
 func newContext() context.Context {
-	return context.Background()
+	return config.WithConfig(context.Background(), cfg)
+}
+
+func exitWithError(err error) {
```
brandonbloom, line 33: see `cmdutil.Fatalf`

config/config.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/BurntSushi/toml"
+)
+
+type Config struct {
+	HomeDir string
+	BinDir  string
+	VarDir  string
+	RunDir  string
+
+	Telemetry struct {
```
brandonbloom, line 17: nit: probably best to avoid anonymous structs like this. Easy enough to make a `TelemetryConfig` type.

config/config.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/BurntSushi/toml"
+)
+
+type Config struct {
+	HomeDir string
```
brandonbloom, line 12: We should be explicit with the `toml:"whatever"` tags, especially since I think underscores as word separators is idiomatic there (right?)

config/config.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	if cfg.BinDir == "" {
+		cfg.BinDir = getExoPath(cfg.HomeDir, "bin", "EXO_BIN")
+	}
+	if cfg.RunDir == "" {
+		cfg.RunDir = getExoPath(cfg.HomeDir, "run", "EXO_RUN")
+	}
+	if cfg.VarDir == "" {
+		cfg.VarDir = getExoPath(cfg.HomeDir, "var", "EXO_VAR")
```
brandonbloom, line 44: Are we expecting to need these extra environment variables if/when we do packages for Homebrew, etc?

config/config.go
```diff
@@ -0,0 +1,73 @@
...
+func MustLoadDefault(cfg *Config) {
+	if err := LoadDefault(cfg); err != nil {
+		panic(err)
+	}
+}
+
+func loadFromFile(filePath string, cfg *Config) error {
+	_, err := toml.DecodeFile(filePath, cfg)
+	if os.IsNotExist(err) {
+		if err = os.WriteFile(filePath, []byte{'\n'}, 0644); err != nil {
```
brandonbloom, line 59: Should we //go:embed an example config file with stuff commented out?

config/config.go
```diff
@@ -0,0 +1,73 @@
...
+			return fmt.Errorf("creating initial config file: %w", err)
```
brandonbloom, line 60: Is this actually an error condition or is it success where cfg is expected to be set to a zero value?

config/config.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+	}
+	return err
+}
+
+func WithConfig(ctx context.Context, cfg *Config) context.Context {
+	return context.WithValue(ctx, configContextKey, cfg)
+}
+
+func GetConfig(ctx context.Context) (*Config, bool) {
```
brandonbloom, line 70: you're not using this anywhere yet, do you expect to? If possible, would be nice to avoid context with/get value things as best we can and be explicit with dependency injection unless we need to cross interface boundaries we don't control.

brandonbloom (commented):
looking pretty good to me so far!

config/config.go
kendru: Haha - I already did this locally :)

config/config.go
kendru: I think this should be an error condition. The zero value *should* be useful, but if we try to use a config file that we can't access, I think that should be an error rather than a silent failure.

config/config.go
kendru: This started as part of the telemetry disabling. Explicit dependency injection would be nice (especially using something like wire), but I think that the context thing is good for now, then we can do a better job segregating things later. I mean, we should probably have a Telemetry struct and pass that in everywhere, but (especially with multiple entry points), it is a lot to wire up manually.

config/config.go
kendru: No, I think that I can just use a flag or environment variable to specify an override location for the config file, then rely on that for any overrides.

config/config.go
kendru: As far as I can tell, there is not really a standard naming convention. I have seen camel, snake, and kebab case. The default w/ the toml library is camelCase, but if you would prefer underscores, I can add explicit tags.

config/config.go
brandonbloom: Yeah, I'd love to _minimize_ the customization and config that is possible :)

config/config.go
brandonbloom: My personal preference is kebab-case, but there are underscores here: https://github.com/toml-lang/toml#example
Also, just generally, I seem to remember underscores being prevalent in INI files, so I think we should go with that.

brandonbloom: Er nevermind, git config seems to use camelCase, so maybe that's fine too.

config/config.go
kendru: Rethinking this. It would not be too much work to do a better job of this. I'll update the PR.

brandonbloom:
Thanks again for correcting this quickly!

PR #63, opened 2021-07-29
Honor disabling of telemetry on config

kendru:
Honor `telemetry.disable` flag in config.


kendru:
This is now incorporated in #62

PR #66, opened 2021-07-30
Linkify log URLs

jwmza:
This splits log messages by spaces and uses anchor tags to display clickable links before stitching them back together.

Should do some due diligence to solve any potential security issues arising from pre-fetching before merging.


jwmza:
Okay, this does two things now:

- Adds basic linkification to logs
- Adds a `/test/` directory to the `gui` which contains pages only for the purposes of testing UI components without relying on actual data or context from real pages. This will give a good place to develop many variants of a single component and focus on front-end functionality and styling, as well as serve as a place to reference how components might be used.
    - As mentioned elsewhere, it would be nice if we could effectively **remove** `/test/` from the repository right when the app is being bundled as these pages should not be concluded in the production app. They should however be part of the git repository so that anybody on the team can use and share test pages.

gui/src/components/logs/FormattedLogMessage.svelte
```diff
@@ -0,0 +1,21 @@
+<script lang="ts">
+  export let message: string = '';
+
+  let words = message.split(' ');
```
brandonbloom, line 4: This is going to create _tons_ of empty spans when there is whitespace in output (which is extremely common!).

A better approach would be to create a parse loop that searches for matching urls. To avoid retesting each element against the regex, you could create a small "AST". Something like `interface Word { link: boolean, value: string }`

gui/src/components/logs/FormattedLogMessage.svelte
```diff
@@ -0,0 +1,21 @@
...
+</script>
+
+<span>
+  {#each words as word}
+    {#if linkRegex.test(word)}
+      <a href={word}>{word}</a>
+    {:else}
+      <span>{word}</span>
+    {/if}
+    {' '}
```
brandonbloom, line 16: This will _always_ include a trailing space. This might not be a problem, but could affect word wrapping or similar. If you do the parse loop solution mentioned above, the correct amount of spaces will be encoded in the spans, so you won't need to manually re-insert it.

jwmza:
Parsing improved but still an inefficient algorithm

jwmza:
Also I'm not sure what precisely what you mean by retesting each element @brandonbloom, assuming you're referring to the Svelte component re-computing the linkified message I'm not sure when/why this happens but if it's an issue then I might suggest this parsing should be moved to the backend somehow such that the Svelte component doesn't need to run significant logic for every message multiple times.

PR #68, opened 2021-07-30
handle a lot more bash syntax in procfiles

brandonbloom:
Replaced ``// TODO: Handle argument quoting.` with a pretty robust bash parser.


brandonbloom:
Resolves argument handling issue as reported by @siggisim

Issue #38, opened 2021-07-27
Pick a better port for syslog server


brandonbloom:
We probably should make this configurable. It looks like 4500 is commonly used for IPSec NAT traversal, so this might break some VPNs.

_Originally posted by @kendru in https://github.com/deref/exo/pull/23#r677047390_

brandonbloom:
assigning to @kendru to address as part of the configuration changes he's making in support of dogfooding.

kendru:
This is now configurable per #76

PR #76, opened 2021-08-03
Run exo with exo

kendru:
Yo dawg. I heard you like exo...

This work will allow us to dev the local version of exo but run it using a released version of exo.


kendru:
Fixes #38

util/strutil/strutil.go
```diff
@@ -0,0 +1,37 @@
+package strutil
```
kendru, line 1: I ended up refactoring some code that used this earlier to not use it. Unless anyone thinks it's worth keeping, I'll toss it out.

providers/unix/components/process/process.go
```diff
@@ -70,29 +75,6 @@ func (p *Process) start(ctx context.Context) error {
 		Setsid: true, // Run in background.
 	}
 
-	// Forward environment.
```
kendru, line 73: Since we are not adding environment variables to send to `supervise`, I *think* that we can just rely on default environment inheritance.

providers/unix/components/process/process.go
```diff
@@ -51,6 +50,11 @@ func (p *Process) start(ctx context.Context) error {
 		gracePeriod = *p.ShutdownGracePeriodSeconds
 	}
 
+	childEnv, err := json.Marshal(p.Environment)
```
kendru, line 53: Yeah, more JSON strings :)

cmd/exo/daemon.go
```diff
@@ -101,8 +101,14 @@ func loadRunState() error {
 }
 
 func newClient() *client.Root {
+	url := cfg.Client.URL
+	if url == "" {
+		url = runState.URL
+	}
+	url += "/_exo/"
```
kendru, line 108: I made the change to include the `/` on the path rather than rely on a trailing slash on the URL itself, since this is more convential.

brandonbloom: That's ok by me, though I think I tried that at some point and backed it out for some reason. May have just been b/c it had knock-on effects and I was already in the middle of another change.

cmd/exo/root.go
```diff
@@ -25,7 +25,7 @@ For more information, see https://exo.deref.io`,
 }
 
 func newContext() context.Context {
-	return config.WithConfig(context.Background(), cfg)
+	return context.Background()
```
brandonbloom, line 28: We're going to conflict here, but it should be a trivial merge.

cmd/logd/main.go
```diff
@@ -30,9 +30,11 @@ func main() {
 	config.MustLoadDefault(cfg)
 	paths := cmdutil.MustMakeDirectories(cfg)
 
-	logd := &logd.Service{}
+	logd := &logd.Service{
```
brandonbloom, line 33: We're going to conflict here too. Sorry!

cmd/logp/main.go
```diff
@@ -25,8 +25,10 @@ func main() {
 	config.MustLoadDefault(cfg)
 	paths := cmdutil.MustMakeDirectories(cfg)
 
-	logd := &logd.Service{}
```
brandonbloom, line 28: same conflict issue with Logger, again: sorry...

config/config.go
```diff
@@ -76,11 +80,29 @@ func loadFromFile(filePath string, cfg *Config) error {
...
+	// Log
+	if cfg.Log.SyslogPort < 1 {
+		cfg.Log.SyslogPort = 4500
+	}
 
-func GetConfig(ctx context.Context) (*Config, bool) {
-	cfg, ok := ctx.Value(configContextKey).(*Config)
-	return cfg, ok
+	// GUI
+	if cfg.GUI.Port < 1 {
```
brandonbloom, line 105: Why < 1 (here and elsewhere?) are negative numbers possible or meaningful in any way?

exod/main.go
```diff
@@ -57,38 +54,39 @@ func RunServer() {
...
-			dumpFile, err := os.OpenFile(dumpPath, os.O_WRONLY|os.O_CREATE|os.O_SYNC|os.O_TRUNC, 0600)
-			if err != nil {
-				golog.Printf("creating %s dump file: %v", redirect.Name, err)
-			}
-			if err := sysutil.Dup2(int(dumpFile.Fd()), redirect.FD); err != nil {
-				golog.Printf("redirecting %s: %v", redirect.Name, err)
-			}
-		}
-	}
+	// TODO: Disable on flag for exo-on-exo mode.
```
brandonbloom, line 57: yeah, this was always kinda janky

exod/main.go
```diff
@@ -109,16 +107,18 @@ func RunServer() {
...
 		Store:      store,
 		Telemetry:  tel,
-		SyslogPort: log.SyslogPort,
+		SyslogPort: cfg.Log.SyslogPort,
 		Docker:     dockerClient,
 	}
 
-	logd := &logd.Service{}
-	logd.VarDir = kernelCfg.VarDir
+	logd := &logd.Service{
```
brandonbloom, line 114: Logger conflict again here

util/strutil/strutil.go
brandonbloom: yeah, these are pretty generic & we should avoid introducing new escaping syntaxes of our own :)

providers/unix/components/process/process.go
brandonbloom: OK, I _think_ I understand this change. Is the idea that you've moved this out of the process starting logic and into the supervisor? But the behavior is otherwise pretty much the same? If so, that seems reasonable to me. I think I did it this way originally to keep the supervisor simpler and less likely to fail, but it has a basic error reporting mechanism, so it should be ok. Tho the more code in there, the more chances for badly handled panics.

Was there something that motivated this change?

providers/unix/components/process/process.go
brandonbloom: I reassigned https://github.com/deref/exo/issues/47 to you. Please improve :)

brandonbloom (commented):
Looks like there are a couple remaining WIP things, but otherwise LGTM.

doc/project-philosophy.md
```diff
@@ -0,0 +1,3 @@
+# Project Philosophy
+
+TODO
```
brandonbloom, line 3: lol yuuuup

providers/unix/components/process/process.go
kendru: The reason for this change is that we were previously adding anything from `p.Environment` to the `supervise` process. In most cases this would be fine, but since we are running `exo`, we need to pass an `EXO_HOME` environment variable. If this variable is loaded by `supervise`, it does the wrong thing (and crashes).

brandonbloom: Aaaah, I gotcha now. That makes sense. To restate: You're delaying the environment application to the actual supervised process, not the supervisor as well. Sounds good 👍🏻

cmd/exo/daemon.go
kendru: I tried things out after the fact, and everything *seemed* to work, but it does make me a bit nervous to change.

brandonbloom: as long as the CLI and the GUI both work, should be fine

config/config.go
kendru: I think that I should change this to be `!= 0` and update the port types in config to `uint` instead.

gui/src/components/VersionInfo.svelte
```diff
@@ -2,6 +2,7 @@
   import { onDestroy, onMount } from 'svelte';
   import { api } from '../lib/api';
   import Button from './Button.svelte';
+import ErrorLabel from './ErrorLabel.svelte';
```
brandonbloom, line 5: do you have prettier running?

PR #79, opened 2021-08-03
dependency injection & internal logging improvements

brandonbloom:
There is increasing complexity and parallelism in the internals of things in support of Docker & I wanted to improve the logs from the exo server itself. I've done the minimum work here to make sure we always use an injected logger. As follow-up work, I'll start to add prefixes or other structured logging data, so that we can correlate messages for a particular request or apply or whatever.

### DIFF
--- a/cmd/exo/run.go
+++ b/cmd/exo/run.go
@@ -2,12 +2,12 @@ package main
 
 import (
 	"fmt"
-	"log"
 	"os"
 	"os/signal"
 
 	"github.com/deref/exo/core/api"
 	"github.com/deref/exo/util/cmdutil"
+	"github.com/deref/exo/util/logging"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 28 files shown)


brandonbloom:
This gives us logs like this:

```
2021/08/03 14:34:29 http 9k2cpnsh353m7rt6vrcyrt9r9c: POST /_exo/workspace/apply?id=nndtq3wrwahjjrhmyaty2ffb34
2021/08/03 14:34:30 http 9k2cpnsh353m7rt6vrcyrt9r9c: status 200 - 246ms
```

And there is a `.Logger` available to all the component lifecycle apis etc.

cmd/exo/run.go
```diff
@@ -33,6 +33,7 @@ If a workspace does not exist, one will be created in the current directory.
 		ctx := newContext()
 		ensureDaemon()
 		cl := newClient()
+		logger := logging.CurrentLogger(ctx)
```
kendru, line 36: Weren't we trying to avoid context-global state?

kendru (commented):
LGTM as long as you are okay with making the logger global and not injecting it explicitly.

cmd/exo/run.go
brandonbloom: Yes, but I needed some way to pass it to HTTP requests. It's explicit pretty much everywhere else & I didn't change `newContext` to something like `newDefaultDependencies()` or whatever to keep the change smaller.

PR #82, opened 2021-08-03
scaffolding for UI that shows networks, volumes, etc

brandonbloom:
assigning to @kendru to review (post-hoc), but tagging @BenElgar and @jwmza for visibility. lots of design work to do here for what we want to appear on these pages

image
image
image


jwmza:
@brandonbloom I'm planning on building the underlying GUI layout (see designs in Slack) for this tonight/tomorrow btw.

Lots of room for modification but I think we know the rough layout, yes? L+T navigation and workspace context

PR #83, opened 2021-08-03
Add process details view

BenElgar:
https://user-images.githubusercontent.com/3443024/128100222-2cf51f6b-979e-421a-82c7-d5633f1999e5.mp4


cmd/exo/status.go
```diff
@@ -17,12 +18,18 @@ func init() {
 	rootCmd.AddCommand(statusCmd)
 }
 
+type ProcessDetails struct {
```
brandonbloom, line 21: unused type?

cmd/exo/status.go
```diff
@@ -17,12 +18,18 @@ func init() {
...
+	Command    string `json:"command"`
+	CreateTime time.Time
+}
+
 var statusCmd = &cobra.Command{
 	Use:   "status",
 	Short: "Report exo status",
 	Long:  `Report exo daemon status.`,
-	Args:  cobra.ExactArgs(0),
+	Args:  cobra.MaximumNArgs(1),
```
brandonbloom, line 30: unintended change?

core/api/workspace.josh.hcl
```diff
@@ -165,17 +171,28 @@ struct "log-description" {
...
   field "provider" "string" {}
-	field "name" "string" {}
+  field "name" "string" {}
+  field "running" "bool" {}
+}
+
+struct "component-status" {
+	field "component-id" "string" {}
 	field "running" "bool" {}
+  field "envVars" "map[string]string" {}
```
brandonbloom, line 190: i'm good with this file using spaces instead of tabs, not sure why it wasn't formatted correctly previously.

but this line and around this line looks messed up now. Is there some standard formatter we can configure somehow?

example/Procfile
```diff
@@ -1,2 +1,3 @@
 echo: socat TCP4-LISTEN:2000,fork EXEC:cat
 tick: ./bin/tick
+spawn: npx sleep infinity
```
brandonbloom, line 3: we have a helper program called `hang` now that does exactly this & saves us an external dependency on npx

gui/src/components/ProcessList.svelte
```diff
@@ -11,7 +11,7 @@
     deleteProcess,
   } from '../lib/process/store';
   import { setLogVisibility, visibleLogsStore } from '../lib/logs/visible-logs';
-  import type { ProcessDescription } from '../lib/process/types';
+  import type { ComponentDetails, } from '../lib/process/types';
```
brandonbloom, line 14: prettier should be running on svelte files

gui/src/components/ProcessList.svelte
```diff
@@ -121,7 +121,7 @@
               {/if}
             </td>
 
-            <td><h2>{name}</h2></td>
+            <td><button on:click={() => void router.push(`#/workspaces/${encodeURIComponent(workspaceId)}/processes/${encodeURIComponent(id)}`)}>{name}</button></td>
```
brandonbloom, line 124: can we use a Link component of some kind here? button seems like both the wrong semantics and appearance.

gui/src/lib/api.ts
```diff
@@ -289,6 +292,18 @@ export const api = (() => {
         await invoke('refresh-all-components');
       },
 
+      async getComponentDetails(): Promise<ComponentDetails[]> {
+        const processes = await this.describeProcesses();
+        return await Promise.all(
```
brandonbloom, line 297: please make `get-component-status` a bulk operation instead of doing `Promise.all` here.

core/api/workspace.josh.hcl
```diff
@@ -165,17 +171,28 @@ struct "log-description" {
...
+	field "CPUPercent" "float64" {}
```
brandonbloom, line 191: fields in these josh files should use `kebab-case`, as that is unambiguous when translating to other cases.

You should add "CPU" to the `inflect.go` file, so you get the correct capitalization in codegen for that.

providers/unix/components/process/component.go
```diff
@@ -23,5 +23,5 @@ type State struct {
 	Pid           int `json:"pid"`
 	// TODO: Store resolved program path & full effective environment.
```
brandonbloom, line 24: this comment now appears to be out of date

gui/src/pages/Process.svelte
```diff
@@ -0,0 +1,213 @@
...
+      refreshAllProcesses(workspace);
+      if (processList.stage === "success") {
+        process = processList.data.filter(p => p.id === processId)[0];
+      }
+      if (process && process.status.running) {
+        cpuPercentages.push(process.status.CPUPercent)
+        if (cpuPercentages.length > 100) {
+          cpuPercentages.shift()
+        }
+        const sparklineSvg = document.querySelector(".sparkline");
```
brandonbloom, line 41: See https://svelte.dev/tutorial/bind-this

gui/src/pages/Process.svelte
```diff
@@ -0,0 +1,213 @@
...
+      }
+    }, 1000);
+  });
+
+  onDestroy(() => {
+    clearInterval(refreshInterval);
+    unsubscribeProcesses();
+  });
+
+// From https://stackoverflow.com/a/14919494
```
brandonbloom, line 58: Please extract to a `<BytesLabel value={numBytesGoHere}/>` component.

gui/src/pages/Process.svelte
```diff
@@ -0,0 +1,213 @@
...
+}
+</script>
+
+<Layout>
+  <section>
+    {#if process}
+      <div>
+        <div id="heading">
+          <button class="back-button" on:click={() => void router.push(`#/workspaces/${encodeURIComponent(workspaceId)}/`)}>
+            🠔 Back
```
brandonbloom, line 98: We have some mechanism for putting back links in the header, tho I think it might change as the layout overall changes. cc @jwmza for guidance.

core/server/workspace.go
```diff
@@ -641,6 +642,87 @@ func (ws *Workspace) RestartComponent(ctx context.Context, input *api.RestartCom
...
+			if err := jsonutil.UnmarshalString(component.State, &state); err != nil {
+				return nil, fmt.Errorf("unmarshalling state: %w", err)
+			}
+
+			status := api.ComponentStatus{
+				ComponentID: id,
+				EnvVars:     state.FullEnvironment,
+			}
+
+			proc, err := psprocess.NewProcess(int32(state.Pid))
```
brandonbloom, line 668: All this stuff that is analyzing unix processes etc should move to the `/providers/unix/components/process` package. If you do that, then you can use the `newController` function to spec and state unmarshaling automatically.

core/api/workspace.josh.hcl
```diff
@@ -165,17 +171,28 @@ struct "log-description" {
...
 struct "process-description" {
-	field "id" "string" {}
+  field "id" "string" {}
   field "provider" "string" {}
-	field "name" "string" {}
+  field "name" "string" {}
+  field "running" "bool" {}
+}
+
+struct "component-status" {
```
brandonbloom, line 187: This structure seems to be really _process_ status, not arbitrary component status. Maybe these fields should be added to `process-description`?

core/api/workspace.josh.hcl
```diff
@@ -60,49 +60,55 @@ interface "workspace" {
 
     output "components" "[]ComponentDescription" {}
   }
-  
+
+  method "get-component-status" {
```
brandonbloom, line 64: I wonder if we should add this method. It seems like instead we should add specs and state to the results of `describe-components`. In that context, component types are dynamic, so spec and state are just strings interpreted by the particular controller for each component type. However, we have a method `describe-processes` which provides a nice decoded representation of process-specific fields.

The reason I think that this is important is because right now `get-component-status` is implemented by querying process state directly. That means that anybody who wants process-specific component state needs to call that API on-demand when it needs that information. It would be better if all this information is serialized in to state on `refresh`. This way we can have generic logic for invalidation, change management, subscriptions, etc. for all this state.

core/api/workspace.go
```diff
@@ -339,6 +352,17 @@ type ProcessDescription struct {
...
 
+type ComponentStatus struct {
+	ComponentID         string            `json:"componentId"`
+	Running             bool              `json:"running"`
+	EnvVars             map[string]string `json:"envVars"`
+	CPUPercent          float64           `json:"CPUPercent"`
+	CreateTime          int64             `json:"createTime"`
+	ResidentMemory      uint64            `json:"residentMemory"`
+	Ports               []uint32          `json:"ports"`
+	ChildrenExecutables []string          `json:"childrenExecutables"`
```
brandonbloom, line 363: Future looking: This is a case where we might care about the distinction between "components" and "resources". A process component may have child processes, which are not quite controlled components, but are resources worth being aware of. Needs more thought...

In the meantime, it's a little weird that we're only capturing the executable name of each child. If any, I'd have expected child pids. In that case, you could imagine creating a process "resource" from a pid.

core/api/workspace.go
```diff
@@ -339,6 +352,17 @@ type ProcessDescription struct {
 	Running  bool   `json:"running"`
 }
 
+type ComponentStatus struct {
+	ComponentID         string            `json:"componentId"`
+	Running             bool              `json:"running"`
+	EnvVars             map[string]string `json:"envVars"`
```
brandonbloom, line 358: We're calling this simply "environment" in many places.

core/api/workspace.josh.hcl
BenElgar: Yes, I fixed the mixed spaces and tabs — and then reintroduced the problem. :facepalm: I did quickly look for a formatter but the only one I found hasn't been maintained in 6 years: https://github.com/fatih/hclfmt.

Will fix this again.

example/Procfile
BenElgar: Ah the point of npx is to fork and create a new child process. I originally had `bash -c "sleep infinity"` but for some reason that didn't work. Can look into this more if you think it's important.

core/server/workspace.go
brandonbloom: To be clear, I think this stuff belongs: https://github.com/deref/exo/blob/fc97c78c4ae0dae695519ab6105551b1dbf9a071/providers/unix/components/process/lifecycle.go#L42

See the above comment about "refresh" too.

gui/src/components/ProcessList.svelte
BenElgar: Ah, my bad. Set up correctly now.

core/api/workspace.go
BenElgar: My thinking is that "environment" is a little too

core/api/workspace.josh.hcl
BenElgar: Ah, yes, I forgot to rename this. I've done that now.

core/api/workspace.josh.hcl
BenElgar: I thought that there was some value in having some distinction between the description of a process and the minute-to-minute status and being able to query them separately. But I then merged the two concepts and it made stuff a lot simpler. So I think you're right.

core/api/workspace.go
BenElgar: Yeah, I don't think that it's the right thing to capture in the long-term but I do think that it provides the most useful information right now.

core/server/workspace.go
BenElgar: I understand moving it to `refresh`, which seems reasonable, but I'm not sure I get the "If you do that, then you can use the newController function to spec and state unmarshaling automatically."

brandonbloom: Sorry, `newController` is a lower-level bit. The really interesting function is:

```go
func (ws *Workspace) control(ctx context.Context, desc api.ComponentDescription, f interface{}) error {
```

There are numerous examples of using it. What it does is instantiate a controller structure, hydrate it with spec and state, execute a method on it, and then save the modified state back to underlying storage.

core/api/workspace.go
brandonbloom: not sure i follow the distinction you are trying to draw

PR #89, opened 2021-08-04
Storage engine (no its not a database. Nothing to see here. Move along) for logs

kendru:
This is just a draft of the storage engine that I had started for logs. It has the concept of tables with schema that can be scanned and filtered by `col = val` but not much else at this point.


brandonbloom:
Just going to close this for now. Can still peek at the branch if it's relevant.

PR #90, opened 2021-08-04
task progress tracking

brandonbloom:
Demo video:

https://www.loom.com/share/23fc91fd0db746b4b32d2614811ea595

Motivated by the fact that Docker operations are potentially _slooow_.

This contains a simple model of "tasks" for tracking progress. There is a "task-store" service, which has describe, create, update, and evict methods. Tasks are arranged into a hierarchy. The root task in a such a hierarchy is called the "job".

Presently, jobs expire (read: all tasks in the job are deleted) when no tasks in that hierarchy has been updated for some time (an hour). Also, tasks are ephemeral, in memory, and so lost if exod restarts. However, the design is such that non-ephemeral tasks are possible in the future.

Work still to be done:

1) a few of the core workspace and kernel operations are expected to be slow, and so should return a job-id. I've only done this for `refresh` so far.
2) The CLI for refresh is polling the jobs endpoint to produce a very simple UI, but needs a layer of polish.
3) I haven't yet created a `/jobs` gui, to do the same as (2). built out of a general progress component, so it can be used in interactive flows of apply, etc too.


core/server/workspace.go
```diff
@@ -145,12 +147,18 @@ func (ws *Workspace) RefreshAllComponents(ctx context.Context, input *api.Refres
 	if err != nil {
 		return nil, err
 	}
-	// TODO: Parallelism.
-	for _, component := range components.Components {
-		if err := ws.refreshComponent(ctx, ws.Store, component); err != nil {
-			// TODO: Error recovery.
-			return nil, fmt.Errorf("refreshing %q: %w", component.Name, err)
+	if err := ws.TaskTracker.RunFunc(ctx, "refresh", func(refreshTask *task.Task) error {
```
brandonbloom, line 150: I will probably replace this usage of `RunFunc` with `Start/Wait` pairs, so that the return value handling is less ugly & makes less use of higher-order programming. This was just my first attempt at getting this to work.

exod/main.go
```diff
@@ -137,11 +146,24 @@ func RunServer(ctx context.Context, flags map[string]string) {
...
 		ctx, shutdown := context.WithCancel(ctx)
 		defer shutdown()
+
 		go func() {
 			if err := logd.Run(ctx); err != nil {
 				cmdutil.Fatalf("log collector error: %w", err)
 			}
 		}()
+
+		go func() {
```
brandonbloom, line 156: Would love to have timer components driven declaratively.

task/runtime.go
```diff
@@ -0,0 +1,99 @@
...
+func (tt *TaskTracker) createTask(ctx context.Context, parent *Task, def Definition) *Task {
+	input := &api.CreateTaskInput{
+		Name: def.Name(),
+	}
+	if parent != nil {
+		input.ParentID = &parent.id
+	}
+	output, err := tt.Store.CreateTask(ctx, input)
+	if err != nil {
+		panic(err) // XXX
```
brandonbloom, line 41: i need to do something about these

task/server/store.go
```diff
@@ -0,0 +1,124 @@
...
+	sto.mx.Lock()
+	defer sto.mx.Unlock()
+
+	expire := chrono.IsoNano(chrono.Now(ctx).Add(-1 * time.Hour))
+
+	// Gather most recent update times for each job.
+	jobs := make(map[string]string)
+	for _, task := range sto.tasks {
+		updated := jobs[task.JobID]
+		if strings.Compare(task.Updated, updated) > 0 {
```
brandonbloom, line 110: i haven't tested this yet, i bet at least one of these comparison operators is wrong :-P

PR #91, opened 2021-08-04
Server watcher

kendru:
This adds a simple file watcher and uses it to restart the dev server on change.

### DIFF
--- a/cmd/watch/main.go
+++ b/cmd/watch/main.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 	"os/signal"
 	"path/filepath"
+	"strings"
 	"syscall"
 	"time"
 
@@ -158,8 +159,20 @@ func main() {
 	}
 
 	if _, ok := cmd.Flags["r"]; ok {
+		var ignores []string
+		if ignoreStr, ok := cmd.Flags["ignore"]; ok {
+			ignores = strings.Split(ignoreStr, ",")
+		}
+
 		if err := filepath.Walk(dir, func(path string, info fs.FileInfo, err error) error {
 			if info.Mode().IsDir() {
+				// TODO: implement something a little more robust than substring match for ignore.
+				for _, ignore := range ignores {
+					if strings.Contains(path, ignore) {
+						return nil
+					}
+				}
+
 				return watcher.Add(path)
 			}
 
--- a/cmd/watch/main.go
+++ b/cmd/watch/main.go
@@ -1,0 +1,208 @@
+package main
+
+import (
+	"fmt"
+	"io/fs"
+	"os"
+	"os/exec"
+	"os/signal"
+	"path/filepath"
+	"syscall"
+	"time"
+
+	"github.com/deref/exo/util/cmdutil"
+	"github.com/fsnotify/fsnotify"
+)
+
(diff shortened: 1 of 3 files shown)


cmd/watch/main.go
```diff
@@ -0,0 +1,221 @@
...
+
+				case _ = <-childStopped:
+					// Child stopped when we were not expecting a restart, so exit.
+					done <- struct{}{}
+				}
+
+			case runStateRestarting:
+				gracefulShutdownTimer := time.NewTimer(time.Second * time.Duration(10))
+				select {
+				case _ = <-childStopped:
```
brandonbloom, line 111: nit: the assignment is not necessary, you can simply write `case <-childStopped:`

cmd/watch/main.go
```diff
@@ -0,0 +1,221 @@
...
+					gracefullyShutdown(child, sig)
+					state = runStateWaitingChildExit
+
+				case _ = <-childStopped:
+					// Child stopped when we were not expecting a restart, so exit.
+					done <- struct{}{}
+				}
+
+			case runStateRestarting:
+				gracefulShutdownTimer := time.NewTimer(time.Second * time.Duration(10))
```
brandonbloom, line 109: since this timer is only used locally here, not looped over or anything like that, I think it is safe to simply do `case <-time.After(...)`, as that will produce a one-element buffered channel and when the timer fires, the sending code will terminate and the channel will get GC'd, no need to call `Stop`. Tho I guess this does cleanup the timer sooner, so maybe it's worth being verbose?

cmd/watch/main.go
```diff
@@ -0,0 +1,221 @@
...
+				select {
+				case _ = <-childStopped:
+					gracefulShutdownTimer.Stop()
+
+				case _ = <-gracefulShutdownTimer.C:
+					child.Process.Kill()
+					// TODO: try to clean up other processes in the process group.
```
brandonbloom, line 116: There is an example of this in the code. You just do `syscall.Kill(-child.Process.Pid)` - we should make like a `osutil.KillProcessGroup` helper or something

cmd/watch/main.go
```diff
@@ -0,0 +1,221 @@
...
+
+			case runStateWaitingChildExit:
+				gracefulShutdownTimer := time.NewTimer(time.Second * time.Duration(10))
+				select {
+				case _ = <-childStopped:
+					gracefulShutdownTimer.Stop()
+
+				case _ = <-gracefulShutdownTimer.C:
+					child.Process.Kill()
+					// TODO: try to clean up other processes in the process group.
```
brandonbloom, line 129: same comment as above

brandonbloom (approved):
LGTM, tho I didn't look too carefully since it's entirely isolated to this extra watch command. Can revisit in lots of detail if/when watching becomes part of exod.

BenElgar (approved):
LGTM though ultimately I think this belongs in exo itself rather than in an external utility :+1:

kendru:
> 
> 
> LGTM though ultimately I think this belongs in exo itself rather than in an external utility 👍

I agree, but I'm not sure where it fits at the moment. Any thoughts?

PR #93, opened 2021-08-05
Add sidebar navigation + assorted UI improvements

jwmza:
image


brandonbloom:
@kendru do we now have some config mechanism we can use for "feature flags"? Would be nice if there was like a global "experimental features" flag or something so that John can land this, but the left-most bar doesn't show up in release builds if for whatever reason we need to cut a release before those views are fleshed out.

kendru:
@brandonbloom I was just thinking about this yesterday... We do not right now, but I'll go ahead and add something now.

jwmza:
Awesome, thanks @kendru

jwmza:
# Summary of changes

- Refactor overall app layout with left vertical sidebar for page navigation in workspaces and improved styling overall in `Layout` and related components
- Simplify and improve `LogsViewer` styling + provide hover focus colors to improve log UX and mitigate passive distraction from timestamps
- Adds **database, home, layers, networking** icons for left nav
- Improve log color generation
- Adapts all non-workspace pages to the new layout with white backgrounds (will be replaced with a common generic page component)
- Adds correct back button functionality across workspace pages (e.g. Process details)
- Adapt `NotFound` and `Offline` pages to layout
- Add basic styling to Docker upcoming component pages -> Networking, Storage, Components
- Provide good styling for `ComponentTable` component with `2n` highlighting for convenient table reading
- Increase vertical scroller width

jwmza:
This PR is getting pretty big and I don't want it to become a bigger mess. Currently it is in good condition to ship (nothing broken, but some repetitive code could use cleaning up and generalization to reusable components)

@brandonbloom or @kendru please review at earliest convenience and let's merge this one before I refactor the layout again around the new no-header, no-footer, big left nav layout we discussed earlier today.

jwmza:
### New page styling improvements
image
image
image
image

gui/src/pages/NewProcess.svelte
```diff
@@ -48,97 +50,98 @@
...
-        }}
-      >
-        script
-      </button>
-      <button
-        class:selected={structured}
-        on:click|preventDefault={() => {
-          if (structured) {
-            return;
+  <WithLeftWorkspaceNav {workspaceId} active="Dashboard">
```
kendru, line 53: I'd suggest making this its own component and including it in `Layout` rather than using it as a higher order component that wraps the page content.

gui/src/pages/Offline.svelte
```diff
@@ -13,16 +13,18 @@
...
-  <div>
-    <h3>
-      <code>exo</code> server is offline
-    </h3>
-    <p class="hint">
-      Run <Code>exo daemon</Code> in your terminal to start.
-    </p>
-  </div>
-</main>
+<Layout>
```
kendru, line 16: I don't think that this should use the `Layout` because there are no actions possible while the server is offline, so navigation, etc. will be broken.

gui/src/components/Layout.svelte
```diff
@@ -56,17 +56,22 @@
   main {
     display: grid;
     grid-auto-flow: row;
-    grid-auto-rows: 48px 1fr auto;
+    grid-auto-rows: max-content 1fr max-content;
```
kendru, line 59: I forgot about max-content :) Thanks!

gui/public/theme.css
```diff
@@ -59,7 +59,7 @@ h1 {
 /* Avoid awful default scrollbars in most OS/Browsers, especially in horizontally scrolling text displays. */
```
kendru: I missed this in an earlier change, but **thank you** for doing custom scrollbars!

gui/src/components/WithLeftWorkspaceNav.svelte
```diff
@@ -0,0 +1,102 @@
...
+  import Storage from './mono/database.svelte';
+  import Networking from './mono/networking.svelte';
+
+  export let workspaceId: string;
+  export let active: string;
+</script>
+
+<div>
+  <aside>
+    <button
```
kendru, line 15: Can we create a `LinkButton` component or similar?

gui/src/components/WithLeftWorkspaceNav.svelte
```diff
@@ -0,0 +1,102 @@
...
+      on:click={() => {
+        router.push(`#/workspaces/${workspaceId}/networking`);
+      }}
+      class={active === 'Networking' ? 'active' : ''}
+      title="Networking"><Networking /></button
+    >
+  </aside>
+  <section>
+    <slot />
```
kendru, line 45: As mentioned elsewhere, I think that this component should _only_ be the sidebar and let the Layout component determine where to use the sidebar and main content.

gui/src/pages/NewProcess.svelte
```diff
@@ -48,97 +50,98 @@
...
+    <form
+      on:submit|preventDefault={async () => {
```
kendru, line 55: I know that this code was already present, but with this much logic in a callback, I'd suggest refactoring it to a function defined in the `script`. Super low-priority though.

gui/src/pages/Process.svelte
```diff
@@ -62,88 +64,81 @@
...
-            </tr>
-            <tr>
-              <td>Children</td>
-              <td>{process.childrenExecutables?.join(', ') ?? 'None'}</td>
-            </tr>
-          </table>
-          <h3>Environment</h3>
-          <table>
-            {#each Object.entries(process.envVars ?? {}) as [name, val] (name)}
+<Layout showBackButton backButtonRoute={workspaceRoute}>
```
kendru, line 67: I'd suggest cleaning up the props a bit so that the back button is displayed iff `backButtonRoute` is supplied rather than having separate props for (en|dis)abling and the route itself. It does require that we always be explicit with the route, but I think that is a good thing.

gui/src/components/WithLeftWorkspaceNav.svelte
```diff
@@ -0,0 +1,102 @@
+<script lang="ts">
+  import * as router from 'svelte-spa-router';
+
+  import Home from './mono/home.svelte';
+  import Layers from './mono/layers.svelte';
+  import Storage from './mono/database.svelte';
+  import Networking from './mono/networking.svelte';
+
+  export let workspaceId: string;
+  export let active: string;
```
kendru, line 10: I would suggest driving the `active` prop by whether the URL matches the link rather than passing in the active link as a prop. I'm guessing that this is why you made it a higher-order component.

kendru (commented):
The main feedback I have is around the component structure of the lefthand nav. Other than that, everything is pretty minor. It looks great!

gui/src/components/WithLeftWorkspaceNav.svelte
jwmza: Yeah, I'm gonna follow up with a code cleanup PR that turns these left navbar menus into reusable components including the buttons having their own component :)

gui/src/components/WithLeftWorkspaceNav.svelte
jwmza: I see what you mean, but I think this causes issues when you want to show the active menu item for a nested page (i.e. the process details page, which is accessed through the main dashboard using the "home" icon, should still highlight the home icon as the "active tab/page")

gui/src/pages/Process.svelte
jwmza: Good idea

gui/src/pages/Offline.svelte
brandonbloom: `Layout` probably needs a more specific name. `PrimaryLayout` or `StandardLayout` or something like that?

jwmza:
@kendru Re: Layout structure / separate components:

What you're suggesting would be better than the way I'm nesting it now but I think there is an even better alternate approach, which is having specialized layout components that wrap the primary layout component.

E.g. `WorkspaceLayout` would use `Layout`, use `LeftNavbar` or something like that, and then simultaneously manage the navbar for the workspace layout as well as having a `slot`.

Then the structure would be, for the in-workspace pages:
```
<WorkspaceLayout ...props>
   ...page
</WorkspaceLayout>
```
as opposed to
```
<Layout  ...props>
  <WithLeftNavBar ...props>
     ...page
  </WithLeftNavBar>
</Layout>
```

---

Re: offline/not found not having layout:

I think they should still have the **base** layout in order to preserve the visual look of the app.

However, the base layout should not include navigation and such, as you mentioned, those won't work in such pages.
Rather I think the layout system should build on progressively more enriched layout components which include things like navigation menus, status items (e.g. git branch and workspace switcher), etc..., whereas the base layout only contains the basic visual layout needed for cross-app consistency.

brandonbloom (approved):
LGTM, all minor stuff. Let's keep rolling forward :)

gui/src/components/WithLeftWorkspaceNav.svelte
kendru: We should use this: https://github.com/ItalyPaleAle/svelte-spa-router#highlight-active-links

kendru:
> What you're suggesting would be better than the way I'm nesting it now but I think there is an even better alternate approach, which is having specialized layout components that wrap the primary layout component.

I definitely agree that we should have multiple layouts.

> I think they should still have the **base** layout in order to preserve the visual look of the app.

What would you suggest should live in the base layout?

jwmza:
> What would you suggest should live in the base layout?

Something like the core underlying DOM layout which is present on all pages, as an analogy, the top frame of a web browser or the base navigation layout and bars of VS Code. Basically what's in `Layout` right now is pretty close.

PR #96, opened 2021-08-05
Add config-based feature flags

kendru:
This adds the ability to set feature flags in `.dev/config.toml` and use the `IfEnabled` component in the UI to determine whether to render something based on whether the flag is set.


gui/src/components/IfEnabled.svelte
```diff
@@ -0,0 +1,15 @@
+<script lang="ts">
+  import { api } from '../lib/api';
+  import { onMount } from 'svelte';
+
+  export let feature: string;
+
+  let enabled: boolean;
+  onMount(async () => {
+    enabled = await api.kernel.isEnabled(feature);
```
kendru, line 9: This queries the server every time the component mounts, but since the server is local, it should not be a problem for now.

gui/src/pages/Sandbox.svelte
```diff
@@ -0,0 +1,19 @@
...
+  import Layout from '../components/Layout.svelte';
+  import IfEnabled from '../components/IfEnabled.svelte';
+</script>
+
+<Layout>
+  <h1>Sandbox</h1>
+
+  <div>
+    <h2>Feature Flags</h2>
+    <IfEnabled feature="sandbox test">
```
kendru, line 11: @jwmza This is just an example of how to use the `IfEnabled` component

.dev/config.toml
```diff
@@ -1,5 +1,8 @@
 httpPort = 4001
 
+[flags]
```
brandonbloom, line 3: Should we call this `features` to distinguish from the more general concept of options=flags?

gui/src/components/IfEnabled.svelte
brandonbloom: I don't love this, as it will create some flicker/pop-in that we're going to just get used to & then won't notice other flicker or pop-in until something ships.

Is there some way we can make these feature flags more static, like the `import.meta.env` stuff?

internal/featureflag/featureflags.go
```diff
@@ -0,0 +1,19 @@
...
+type FeatureFlags interface {
+	// CheckEnabled takes a list of flags and returns a parallel list
+	// of booleans indicating whether the `i`th flag is enabled.
+	CheckEnabled(ctx context.Context, flags []string) ([]bool, error)
+}
+
+func IsEnabled(ctx context.Context, ff FeatureFlags, flag string) bool {
+	enabled, err := ff.CheckEnabled(ctx, []string{flag})
+	if err != nil {
+		// TODO: What is the right thing to do here?
```
brandonbloom, line 14: At the very least, log something.

gui/src/pages/Sandbox.svelte
```diff
@@ -0,0 +1,19 @@
+<script lang="ts">
```
brandonbloom, line 1: Instead of pages/Sandbox, can you please use storybook?

brandonbloom (commented):
Overall: LGTM. I'd vote to merge in order to unblock John's navigation work on main. However, I think all of my comments require some follow-up.

gui/src/pages/Sandbox.svelte
jwmza: LGTM, thanks :)

gui/src/components/IfEnabled.svelte
kendru: Yeah, I'll make this better.

gui/src/routes.ts
```diff
@@ -16,6 +16,6 @@ export default {
   '/workspaces/:workspace/storage': WorkspaceStorage,
   '/workspaces/:workspace/networking': WorkspaceNetworking,
   '/workspaces/:workspace/new-process': NewProcess,
-  '/workspaces/:workspace/processes/:process': Process,
+  '/workspaces/:workspace/processes/:process': Process,
```
brandonbloom, line 19: is prettier running? :-P

PR #99, opened 2021-08-06
Filter logs

### LINKED ISSUE #94: [FEATURE] Filterable logs
When I have a large volume of logs, and I am looking for a particular string, I cannot do this easily with exo.

I would like the log viewer to include a filter button. Clicking on this button should bring up a text box that I can type in to filter the logs. As I type, the logs are filtered to only those that contain the string that I have entered (case insensitive).

**Describe alternatives you've considered**
Go back to logging to a file then searching with `grep`. 😢

kendru:
https://www.loom.com/share/d984903afb9343e38d5322c69a92f1e1

This is just hacked in for now because my plan is to revisit the storage engine and general state management for logs right after we get the Docker Compose release wrapped up.

Also addresses:
- Wrapping text in long log lines
- GUI breakage when re-mounting the log viewer.

Fixes #94

### DIFF
--- a/gui/src/components/LogsViewer.svelte
+++ b/gui/src/components/LogsViewer.svelte
@@ -3,11 +3,17 @@
   import { onMount, onDestroy, afterUpdate, beforeUpdate } from 'svelte';
   import { hasData, isUnresolved } from '../lib/api';
   import type { WorkspaceApi } from '../lib/api';
-  import { loadInitialLogs, logsStore, refreshLogs } from '../lib/logs/store';
+  import {
+    loadInitialLogs,
+    logsStore,
+    refreshLogs,
+    setFilterStr,
+  } from '../lib/logs/store';
   import type { WorkspaceState } from '../lib/logs/store';
   import { shortDate } from '../lib/time';
   import { processes } from '../lib/process/store';
   import FormattedLogMessage from './logs/FormattedLogMessage.svelte';
+  import debounce from '../lib/debounce';
 
   export let workspace: WorkspaceApi;
   export let workspaceId: string;
@@ -16,6 +22,7 @@
 
   let state: WorkspaceState = {
     cursor: null,
+    filterStr: null,
     events: { stage: 'success', data: [] },
   };
   const unsubscribeLogStore = logsStore.subscribe((workspaces) => {
@@ -81,6 +88,10 @@
       logViewport.scrollTop = logViewport.scrollHeight;
     }
   });
+
+  const setFilterStrDebounced = debounce((fitlerStr: string) => {
+    setFilterStr(workspaceId, workspace, fitlerStr);
+  }, 250);
 </script>
 
 <section>
@@ -99,6 +110,14 @@
         {/each}
       </div>
     </div>
+    <input
+      placeholder="Filter..."
+      value={state.filterStr || ''}
+      on:input={(e) => {
+        const text = e.currentTarget.value.trim();
+        setFilterStrDebounced(text);
+      }}
+    />
   {:else if isUnresolved(state.events)}
     <div>Loading logs...</div>
   {:else}
--- a/gui/src/components/LogsViewer.svelte
+++ b/gui/src/components/LogsViewer.svelte
@@ -164,7 +164,7 @@
 
   td {
     padding: 0 0.3em;
-    white-space: pre;
+    vertical-align: text-top;
     color: #333333;
   }
 
--- a/gui/src/components/LogsViewer.svelte
+++ b/gui/src/components/LogsViewer.svelte
@@ -86,19 +86,17 @@
 <section>
   <h1>Logs</h1>
   {#if hasData(state.events)}
-    <div class="log-table-overflow-wrapper">
-      <div class="log-table-container" bind:this={logViewport}>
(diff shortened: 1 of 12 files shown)


gui/src/components/LogsViewer.svelte
```diff
@@ -81,26 +88,36 @@
...
-              <td>{friendlyName(event.log)}</td>
-              <td>
-                <FormattedLogMessage message={event.message} />
-              </td>
-            </tr>
-          {/each}
-        </table>
+    <div class="log-overflow-wrapper">
+      <div class="log-container" bind:this={logViewport}>
+        {#each state.events.data as event (event.id)}
```
jwmza, line 102: Why did this change from `<table>` to styled flexbox `<div>` structure?

The reason this was originally changed to a table is because it automatically aligns all the timestamps and colored process titles vertically, which I figure is important, and as far as I can tell this new structure would make those titles misaligned if they have different lengths.

In theory, we *can* get the same effect without `<table>`, by doing the following:

- Using the list of all processes which have `logs=visible`, pad the right hand side of their titles with enough spaces to make all of their title strings the same number of characters long (i.e. the length of the longest one of them all) in the `LogsViewer`. Since logs are monospaced this should work.
- In a similar process, pad and clamp the timestamp strings up to a certain length, so that they are all the same length (I believe they already are, but I'm not sure if the current formatting is guaranteed to do that)

gui/src/components/LogsViewer.svelte
```diff
@@ -81,26 +88,36 @@
...
+            <div class="message">
+              <FormattedLogMessage message={event.message} />
+            </div>
+          </div>
+        {/each}
       </div>
     </div>
+    <input
+      placeholder="Filter..."
+      value={state.filterStr || ''}
```
jwmza, line 115: Does this work like a controlled input in React? I'm shocked that this seems to condense
```
value={stateX}
onChange={(e) => setStateX(e.currentTarget.value)}
```
into the far more concise
```
value={state.X || ''}
```
if so. Pretty cool, but it feels unintuitive that we don't have to explicitly call a change to the state.

jwmza (dismissed):
The Loom demo & code looks great - loving the way this looks. Already seems like it'd be super valuable for debugging and sifting through tons of logs to find something particular.

Please address the issue I mentioned with the horizontal alignment of the log title labels, though - the move from `<table>` to nested flex `<div>`.

Other than that, LGTM

jwmza:
@kendru I guess the filter input also needs some styling - you could try adapting the ones from the Add Process flow if you'd like (I think we actually have a wrapped input component in `components`) or leave it for me to follow up with a styling pass later if you'd like.

gui/src/components/LogsViewer.svelte
kendru: `value` here works like `defaultValue` in React, which is fine in this case because I just care about updating the external state, not synchronizing when someone else changes that state. To do the 2-way binding you are talking about is even simpler:

```
<script>
let x = '';
</script>

<input bind:value={x} />
```

kendru:
@jwmza I think it would be great if you took a styling pass when you have time because even if I do "real" styling, you would probably still need to take another pass for it to actually look good ;) If you do not have time, let me know, and I can at least make it a little less ugly.

jwmza:
> @jwmza I think it would be great if you took a styling pass when you have time because even if I do "real" styling, you would probably still need to take another pass for it to actually look good ;) If you do not have time, let me know, and I can at least make it a little less ugly.

I'm more than happy to - will probably be either Sunday or Monday though, but this should be fine as we aren't releasing until after then.

gui/src/components/LogsViewer.svelte
kendru: Yeah - you are right - they do end up misaligned. I ended up changing this to try to get the text wrapping working with the process label only displaying on the first line, but I changed that design decision and never backed out the flexbox stuff. I think that we may eventually want to go more fixed-width with the log name since the filtering means that as you filter, the column width can change based on the longest log name of the log events that match that filter. I'll back this change out and just do the line wrapping within the table (which I'm assuming is simple as removing the `whitespace: pre`).

internal/core/api/workspace.josh.hcl
```diff
@@ -109,6 +109,7 @@ interface "workspace" {
     input "logs" "[]string" {}
 
     input "cursor" "*string" {}
+    input "filterStr" "*string" {}
```
brandonbloom, line 112: Seems like null/undefined and empty string have the same behavior, so we can probably just omit the `*`. We need some way to encode this conveniently for producing the right `filterStr?: string` in the typescript code, but like provide a hint or overlay or something or driven by `minLength` or something like that for Go.

jwmza:
@kendru I don't think removing `whitespace: pre` would work but I'm not sure what the inherited default would be. Maybe it does.

Instead what I would recommend trying is use the same `table` setup we had before, but use

```
td {
  whitespace: pre-wrap;
}
```

I believe I tested this earlier today and it works as desired. Lmk if not

Pre-wrap should simultaneously maintain the whitespace formatting of the raw data and also wrap the lines if they exceed the size of the container

PR #100, opened 2021-08-06
Pass JSON config to supervise on stdin

### LINKED ISSUE #47: pass config to supervise process on stdin

kendru:
Fixes #47


cmd/supervise/main.go
```diff
@@ -1,13 +1,21 @@
...
 	"github.com/deref/exo/internal/supervise"
+	"github.com/deref/exo/internal/util/cmdutil"
 )
 
 func main() {
-	command := os.Args[0]
-	args := os.Args[1:]
-	supervise.Main(command, args)
+	cfg := &supervise.Config{}
+	if err := json.NewDecoder(os.Stdin).Decode(cfg); err != nil {
```
brandonbloom, line 13: why duplicate this code rather than stick it inside `supervise.Main`?

internal/providers/unix/components/process/process.go
```diff
@@ -47,30 +48,24 @@ func (p *Process) start(ctx context.Context) error {
...
-		p.Arguments...,
-	)
-	cmd := exec.Command(supervisePath, superviseArgs...)
+	cmd := exec.Command(supervisePath, "supervise")
 	cmd.SysProcAttr = &syscall.SysProcAttr{
 		Setsid: true, // Run in background.
 	}
 
+	// Pipe JSON config to supervise on stdin.
+	configJSON := supervise.MustEncodeConfig(&supervise.Config{
```
brandonbloom, line 59: I don't see this function defined anywhere...

brandonbloom (changes requested):
I _think_ this PR is missing a file or something for the MustEncodeConfig function?

cmd/supervise/main.go
kendru: I don't want supervise.Main to depend on reading from `os.Stdin`, and it is called from exop and exod, which do not expect input on stdin.

internal/providers/unix/components/process/process.go
kendru: I just added the file that *should* have been there initially. Thanks for catching this!

cmd/exop/main.go
```diff
@@ -30,15 +29,24 @@ func main() {
 	// subcommand so that it always uses the latest veresion of the code rather than relying
```
brandonbloom, line 29: typo: veresion

cmd/supervise/main.go
brandonbloom: I'm confused about this. Seems like there is some leftover dev stuff that is marked "XXX" and "This is broken". Should we remove/cleanup that code?

Regarding `exop` - I'm not sure why that code would need a `supervise` codepath at all. I don't remember adding that. There is already the `supervise/main.go` codepath for testing the supervise script directly without the exo binary. Integrating it in to exop seems redundant.

As for the exod codepath, I think that was a quick and dirty hack to deal with the fact that exod wants to launch supervise by looking for `argv[0]` to spawn a new process. Now that we have a proper config mechanism, maybe we should use that instead to supply a supervise path?

brandonbloom (approved):
LGTM. I might take the opportunity to do one more cleanup step improving, removing or otherwise commenting the various entrypoints. Up to you.

kendru:
@brandonbloom I removed the entrypoints to supervise from `exop` and `exod`, and I removed the top-level `supervise` command entirely. If you use exop and exod and need something from them, let me know and I can add it.

PR #103, opened 2021-08-06
Add dexo

BenElgar:


kendru:
Is there a reason to not just run `exo run` from the repo root?

BenElgar:
The point of this was to allow me to have `exo` always point to the prod version of exo I have installed and `dexo` to point to the development one. This way I can do `cd node-js-getting-started && dexo run Procfile` or whatever and have it use the latest version of the exo binary to issue a command against the development server. How would you do the equivalent (without constantly tweaking `~/.exo/config.toml`)?

script/dexo
```diff
@@ -0,0 +1,23 @@
...
+# run ./cmd/exo" in development. It builds the version of exo on HEAD and
+# configures the client to talk to the development instance of exo and can be
+# run from any working directory.
+
+script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+
+pushd "$script_dir/.." > /dev/null
+
+tmp_dir=$(mktemp -d)
+go build -o "$tmp_dir" ./cmd/exo
```
kendru, line 14: You could put the binary in `.dev` if you wanted, since everything in that directory is untracked.

kendru:
@BenElgar Ah, that makes sense now - that this is just for CLI use. Looks good to me!

PR #106, opened 2021-08-06
Add storybook to procfile, make "npm i" execution manual.

brandonbloom:
dev workflow impact, so FYI everybody: @BenElgar @kendru @jwmza


jwmza:
Bonus: Adds one more process in the exo² GUI am I right? 😄

PR #114, opened 2021-08-09
Improve filter input styling

jwmza:
Adds similar styling to the log viewer filter input as the new process form text inputs have, as an improvement over the (inconsistent) browser defaults.

image

This uses Svelte's `--style-props` to allow overriding of the default `300px` `<Textbox />` input width without complicating the code for the actual components.

---

- This also simplifies the styling for the dev footer (note that the footer is going away / being replaced somewhere else in ~1 day or so as we swap to a left-sidebar-only layout with no global footer or header)

image

### DIFF
--- a/gui/package-lock.json
+++ b/gui/package-lock.json
@@ -4784,6 +4784,16 @@
       "integrity": "sha1-CvMqmm4Tyno/1QaeYtew9Y0NiUY=",
       "dev": true
     },
+    "babel-plugin-transform-vite-meta-env": {
+      "version": "1.0.3",
+      "resolved": "https://registry.npmjs.org/babel-plugin-transform-vite-meta-env/-/babel-plugin-transform-vite-meta-env-1.0.3.tgz",
+      "integrity": "sha512-eyfuDEXrMu667TQpmctHeTlJrZA6jXYHyEJFjcM0yEa60LS/LXlOg2PBbMb8DVS+V9CnTj/j9itdlDVMcY2zEg==",
+      "dev": true,
+      "requires": {
+        "@babel/runtime": "^7.13.9",
+        "@types/babel__core": "^7.1.12"
+      }
+    },
     "babel-preset-current-node-syntax": {
       "version": "1.0.1",
       "resolved": "https://registry.npmjs.org/babel-preset-current-node-syntax/-/babel-preset-current-node-syntax-1.0.1.tgz",
--- a/gui/public/theme.css
+++ b/gui/public/theme.css
@@ -1,3 +1,11 @@
+/* Default global styles for component properties that can be overridden with `style-props` */
+
+:root {
+  --input-width: 300px;
+}
+
+/* Theme color variables */
+
 :root {
   --text-color: #061a29;
   --bg-color: #fff;
--- a/gui/src/components/Layout.svelte
+++ b/gui/src/components/Layout.svelte
@@ -99,16 +99,13 @@
     padding: 4px 6px;
     height: 20px;
     align-items: center;
+    font-size: 12px;
     background: #eeeeee;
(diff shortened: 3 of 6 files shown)


brandonbloom:
Instead of styling the `input`, change the code to use the `Textbox` component we already have.

I think that `className` props and the `{...somehow pass ALL props...}` stuff is a bit of a code smell. It _can_ be done, but it really should be an allow-list, otherwise you can have a conflict/breaking-change when somebody _adds_ a property, which should never happen. Better is to make a narrow interface with just the props you need. This can unfortunately be difficult for layout but if you have specific issues, we can walk through various approaches together.

jwmza:
Okay, I think I figured out an approach that works fairly well for this, using Svelte's `--style-props`.

However, I am now running into a few issues with this filter input in general - please have a look @kendru 

(a) I am getting this Typescript issue: `Property 'value' does not exist on type 'EventTarget'.ts(2339)` at this code
```
    <Textbox
      placeholder="Filter..."
      value={state.filterStr || ''}
      on:input={(e) => {
>>      const text = e.currentTarget?.value.trim();    <<
        setFilterStrDebounced(text);
      }}
      --input-width="100%"
    />
```
tried to fix it but to no avail. My simplest attempt which I hoped would work would be to do a messy type cast thing like `(e.currentTarget as any)` or something like that, but that doesn't seem to work in these Svelte event handler script blocks.

(b) There is a lot of flickering/strange behavior around the contents of the input here and I think it has to do with the way Svelte is handling `value` on the input element in combination with the debounced `state.filterStr`. Instead I can imagine it might be good to have an instantly bound input value variable which is 100% reactive and never flickers/debounces, and then debounce a second underlying variable which is used in the filtering itself. Maybe we could pair on this change @kendru ?

jwmza:
I'm still not caught up on all of the best practices in Svelte components so I may be missing a few things here

brandonbloom:
> --style-props

I was unaware of both that, and it's underlying browser mechanism `display: contents`. Both seem very useful! Glad you found that.

> My simplest attempt which I hoped would work would be to do a messy type cast thing like (e.currentTarget as any)

I think that TypeScript syntax is not allowed in the HTML portion of a svelte file. The simple work around is to make a `handleWhatever` function in the `<script lang="ts">` section, though I imagine this is a common complaint. Can look to see if their is a solution somewhere.

You should be able to ignore the typescript error for now, as we have various typescript errors leftover from when we didn't have strict=true enabled. We'll fix them in time.

I'll let @kendru respond regarding binding & debouncing. The UI state should not be debounced, but the filter state can be. Though it might be better to have an explicit state machine rather than encoding the debouncing state in to callbacks and timeouts. My experience is that implicit state in UI leads to subtle timing bugs.

brandonbloom (approved):

jwmza:
> I think that TypeScript syntax is not allowed in the HTML portion of a svelte file. The simple work around is to make a handleWhatever function in the <script lang="ts"> section, though I imagine this is a common complaint. Can look to see if their is a solution somewhere.

As fate would have it, I did make a `handleInput` function in `<script lang="ts">` and it caused a similar TS issue because it was no longer happy with `e` not satisfying the specific requirements of `on:input` handlers (it wanted `Event`, not `InputEvent`)

jwmza:
> The UI state should not be debounced, but the filter state can be. Though it might be better to have an explicit state machine rather than encoding the debouncing state in to callbacks and timeouts. My experience is that implicit state in UI leads to subtle timing bugs.

I agree w/ this

For now, I'll merge this PR as it specifically addresses styling, and hopefully @kendru and I can work together tomorrow or soon this week on a follow-up regarding the state/debouncing.

PR #115, opened 2021-08-09
Use localhost for syslog address

### LINKED ISSUE #110: [BUG] Doesn't work on linux because of host.internal.docker
**Describe the bug**
host.internal.docker is a host that is added on both Windows and MacOS but doesn't exist in Linux. Exo uses this host for the syslog-address.

We need to do something like this: https://stackoverflow.com/a/62431165

BenElgar:
I spent ages trying to debug DNS resolution in the container but I realised that the issue appears to be that the docker daemon on the host is actually handling syslog forwarding so we just need to use "localhost".

Could someone on a MacOS/Windows environment confirm this works?

Fixes #110

### DIFF
--- a/internal/providers/docker/components/container/lifecycle.go
+++ b/internal/providers/docker/components/container/lifecycle.go
@@ -79,7 +79,7 @@ func (c *Container) create(ctx context.Context) error {
 		// syslog service.
 		logCfg.Type = "syslog"
 		logCfg.Config = map[string]string{
-			"syslog-address":  fmt.Sprintf("udp://host.docker.internal:%d", c.SyslogPort),
+			"syslog-address":  fmt.Sprintf("udp://localhost:%d", c.SyslogPort),
 			"syslog-facility": "1", // "user-level messages"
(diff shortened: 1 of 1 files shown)


brandonbloom:
I could have sworn I tried this and it didn't work, but I'll try again.

@kendru please try on Windows.

kendru:
Since I am running in WSL, it is the same experience as Linux for me. This looks correct, since the syslog address should be what the _docker daemon_ resolves, not the containerized process.

brandonbloom:
Except that Mac (unlike Linus and WSL) doesn't have real containerization & so all sorts of weirdness happens. For example, folks have had problems with `localhost` resolving and had to use `127.0.0.1` for syslog: https://www.reddit.com/r/docker/comments/m6ykrr/docker_to_syslog_stopped_working/

I haven't gotten a chance to verify this, but again, I'm almost certain that neither localhost nor 127.0.0.1 worked correctly for me. I had to go out of my way to find out that `host.docker.internal` existed.

PR #117, opened 2021-08-09
Record image ID from docker build

BenElgar:


internal/providers/docker/components/container/build.go
```diff
@@ -91,13 +94,39 @@ func (c *Container) buildImage(ctx context.Context) error {
...
+						Code    int    `json:"code"`
+						Message string `json:"message"`
+					} `json:"errorDetail"`
+					Aux struct {
+						ID string `json:"ID"`
+					} `json:"aux"`
+				}
+				err := json.Unmarshal(scanner.Bytes(), &d)
+				if err != nil {
+					return err
```
brandonbloom, line 113: Two Go style nitpicks here:

1) Only using `err` in this very next condition, so better to do `if err := ...; err != nil {` idiom.
2) `if err != nil { return err }` should be extremely rare. It's almost always better to add context from the call site. In this specific case, json unmarshaling errors (like most of the stdlib and well behaved code) doesn't mention json on marshalling at all, so you should add context before returning the error. Something like `return fmt.Errorf("unmarshalling response: %w", err)` or something like that is good enough.

internal/providers/docker/components/container/build.go
```diff
@@ -91,13 +94,39 @@ func (c *Container) buildImage(ctx context.Context) error {
...
+				}
+				if d.ErrorDetail.Message != "" {
+					return fmt.Errorf("docker build error: " + d.ErrorDetail.Message)
+				}
+				if strings.HasPrefix(d.Aux.ID, "sha256:") {
+					c.ImageID = d.Aux.ID
+					c.Image = d.Aux.ID
```
brandonbloom, line 120: This is `c.Spec.Image`. The spec should not be mutated as part of the controller's behavior, as those changes are not saved anywhere & are expected to be managed by the user's manifest. Maybe we should start being explicit about `c.State.Whatever` to clarify.

brandonbloom (commented):
minor comments

internal/providers/docker/components/container/build.go
BenElgar: Ah, that makes sense. I think I agree that giving those fields explicit names would be better.

PR #118, opened 2021-08-09
Add telemetry around operation execution

kendru:
This PR adds session tracking for each `exo` daemon run and aggregate statistics for any method invoked. The telemetry requests are proxied through `https://exo.deref.io/api/telemetry` (see [the api function](https://github.com/deref/exo-website/blob/main/pages/api/telemetry.ts)) to Hasura, where they are stored in the database with an association to the exo session to which they belong. See below for an example payload. All time-based stats are in microseconds.

image

### DIFF
--- a/internal/josh/server/methods.go
+++ b/internal/josh/server/methods.go
@@ -15,9 +15,8 @@ import (
 )
 
 type MethodHandler struct {
-	Factory   func(req *http.Request) interface{}
-	Name      string
-	Telemetry telemetry.Telemetry
+	Factory func(req *http.Request) interface{}
+	Name    string
 }
 
 const expectedSignature = "expected signature: func (ctx context.Context, input *YourInput) (*YourOutput, error)"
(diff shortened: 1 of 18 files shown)


internal/josh/server/methods.go
```diff
@@ -58,21 +72,24 @@ func (handler *MethodHandler) ServeHTTP(w http.ResponseWriter, req *http.Request
...
 }
 
 type MuxBuilder struct {
-	prefix  string
-	mux     *http.ServeMux
-	methods []string
+	prefix    string
+	mux       *http.ServeMux
+	methods   []string
+	telemetry telemetry.Telemetry
```
brandonbloom, line 85: becomes part of context

internal/telemetry/default.go
```diff
@@ -77,3 +152,33 @@ func (t *defaultTelemetry) getLatestVersion() (interface{}, error) {
 
 	return string(body), nil
 }
+
+func (t *defaultTelemetry) sendRecordedTelemetry() {
+	// There is the possibility of dropping data since we are not atomically
```
brandonbloom, line 157: not great, b/c this will create noise in the race detector output

kendru: Fixed

PR #122, opened 2021-08-10
Improve process detail table styling

jwmza:
Adds similar styling as the docker component read page tables to the process details tables. Should abstract this table styling to a generic underlying component before merging.

image


BenElgar:
LGTM though having 4 different background colours for table cells feels a little busy to me. Nbd.
image

jwmza:
> LGTM though having 4 different background colours for table cells feels a little busy to me. Nbd.

The idea there is to allow easy scanning of rows and columns alike, but since there are relatively few columns in these tables I could remove that and make it a simple 2-color row striping @BenElgar - let me know if you'd like that change made.

PR #129, opened 2021-08-10
Fix stop components api

BenElgar:


brandonbloom:
Whoops. Sorry. Thanks for fixing this!

brandonbloom:
Actually, I don't think this (or the start-components one) are fully fixed. The arg is now "refs" plural. So `refs: [ref]` here.

PR #130, opened 2021-08-10
Add basic container details

BenElgar:
Adds basic container details to reach parity with unix processes

### DIFF
--- a/internal/core/server/dockerstats.go
+++ b/internal/core/server/dockerstats.go
@@ -1,0 +1,99 @@
+package server
+
+type DockerStats struct {
+	BlkioStats struct {
+		IoServiceBytesRecursive []struct {
+			Major int64  `json:"major"`
+			Minor int64  `json:"minor"`
+			Op    string `json:"op"`
+			Value int64  `json:"value"`
+		} `json:"io_service_bytes_recursive"`
+	} `json:"blkio_stats"`
+	CPUStats struct {
+		CPUUsage struct {
+			TotalUsage        uint64 `json:"total_usage"`
+			UsageInKernelmode uint64 `json:"usage_in_kernelmode"`
+			UsageInUsermode   uint64 `json:"usage_in_usermode"`
+		} `json:"cpu_usage"`
+		OnlineCpus     int64 `json:"online_cpus"`
+		SystemCPUUsage int64 `json:"system_cpu_usage"`
+		ThrottlingData struct {
+			Periods          int64 `json:"periods"`
+			ThrottledPeriods int64 `json:"throttled_periods"`
+			ThrottledTime    int64 `json:"throttled_time"`
+		} `json:"throttling_data"`
+	} `json:"cpu_stats"`
+	ID          string `json:"id"`
+	MemoryStats struct {
+		Limit int64 `json:"limit"`
+		Stats struct {
+			ActiveAnon            int64 `json:"active_anon"`
+			ActiveFile            int64 `json:"active_file"`
+			Anon                  int64 `json:"anon"`
+			AnonThp               int64 `json:"anon_thp"`
+			File                  int64 `json:"file"`
+			FileDirty             int64 `json:"file_dirty"`
+			FileMapped            int64 `json:"file_mapped"`
+			FileWriteback         int64 `json:"file_writeback"`
+			InactiveAnon          int64 `json:"inactive_anon"`
+			InactiveFile          int64 `json:"inactive_file"`
+			KernelStack           int64 `json:"kernel_stack"`
+			Pgactivate            int64 `json:"pgactivate"`
+			Pgdeactivate          int64 `json:"pgdeactivate"`
+			Pgfault               int64 `json:"pgfault"`
+			Pglazyfree            int64 `json:"pglazyfree"`
+			Pglazyfreed           int64 `json:"pglazyfreed"`
+			Pgmajfault            int64 `json:"pgmajfault"`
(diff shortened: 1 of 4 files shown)


internal/core/server/dockerstats.go
```diff
@@ -0,0 +1,99 @@
+package server
+
+type DockerStats struct {
```
brandonbloom, line 3: Can we move this to the `providers/docker` directory and rename it to something like `ContainerStats`?

internal/core/server/workspace.go
```diff
@@ -692,9 +695,7 @@ func (ws *Workspace) DescribeProcesses(ctx context.Context, input *api.DescribeP
 			}
 			output.Processes = append(output.Processes, process)
 		case "container":
```
brandonbloom, line 697: same general feedback as a previous PR: let's move this logic to the respective provider subpackages, even if we hard code dispatch to that logic from here for now.

internal/core/server/workspace.go
```diff
@@ -704,8 +705,71 @@ func (ws *Workspace) DescribeProcesses(ctx context.Context, input *api.DescribeP
 				ID:       component.ID,
 				Name:     component.Name,
 				Provider: "docker",
-				Running:  state.Running,
 			}
+
+			containerInfo, err := ws.Docker.ContainerInspect(ctx, state.ContainerID)
+			if err == nil {
```
brandonbloom, line 711: In general, we should prefer early returns. If this becomes a separate function like `GetContainerStats`, then you can do `if err != nil { return ...`, which outdents all the code a layer.

internal/core/server/workspace.go
```diff
@@ -704,8 +705,71 @@ func (ws *Workspace) DescribeProcesses(ctx context.Context, input *api.DescribeP
...
+				process.Running = containerInfo.State.Running
+				if process.Running {
+					statRequest, err := ws.Docker.ContainerStats(ctx, state.ContainerID, false)
+					if err != nil {
+						return nil, fmt.Errorf("could not get container stats: %w", err)
+					}
+					defer statRequest.Body.Close()
+
+					scanner := bufio.NewScanner(statRequest.Body)
```
brandonbloom, line 720: Why are you using a scanner here? It's a single JSON payload response, right? You can use `json.NewDecoder(statRequest.Body)`

internal/core/server/workspace.go
```diff
@@ -704,8 +705,71 @@ func (ws *Workspace) DescribeProcesses(ctx context.Context, input *api.DescribeP
...
+						return nil, fmt.Errorf("error scanning container stats: %w", err)
+					}
+
+					if !didScan {
+						return nil, fmt.Errorf("could not read container stats")
+					}
+
+					var containerStats DockerStats
+					err = json.Unmarshal(scanner.Bytes(), &containerStats)
+					if err != nil {
```
brandonbloom, line 732: nit: when possible, reduce scope of newly introduced err variable with `if err := json.Unmarshal(...); err != nil {` When I see `err =` without `:`, I go looking elsewhere to see how these errors are handled.

internal/core/server/workspace.go
```diff
@@ -704,8 +705,71 @@ func (ws *Workspace) DescribeProcesses(ctx context.Context, input *api.DescribeP
...
+					process.CreateTime = startTime.UnixNano() / 1e6
+
+					if containerStats.CPUStats.SystemCPUUsage != 0 {
+						process.CPUPercent = float64(containerStats.CPUStats.CPUUsage.TotalUsage) / 1e9
+					}
+
+					process.EnvVars = map[string]string{}
+					for _, env := range containerInfo.Config.Env {
+						decomposedEnv := strings.SplitN(env, "=", 2)
+						process.EnvVars[decomposedEnv[0]] = decomposedEnv[1]
```
brandonbloom, line 752: This will panic if `Config.Env` doesn't contain an `=` for whatever reason. Is that OK?

internal/core/server/workspace.go
```diff
@@ -704,8 +705,71 @@ func (ws *Workspace) DescribeProcesses(ctx context.Context, input *api.DescribeP
...
+					}
+
+					process.Ports = []uint32{}
```
brandonbloom, line 755: When you have a known array length that you're copying/mapping, use this pattern:

```go
process.Ports = make([]uint32, len(containerInfo.NetworkSettings.Ports)
for i, p := range containerInfo.NetworkSettings.Ports {
  processPorts[i] = ...
}
```

internal/core/server/workspace.go
```diff
@@ -704,8 +705,71 @@ func (ws *Workspace) DescribeProcesses(ctx context.Context, input *api.DescribeP
...
+						process.Ports = append(process.Ports, uint32(p.Int()))
+					}
+
+					topBody, err := ws.Docker.ContainerTop(ctx, state.ContainerID, []string{})
+					if err != nil {
+						return nil, fmt.Errorf("could not top container: %w", err)
+					}
+
+					process.ChildrenExecutables = []string{}
+					for _, proc := range topBody.Processes {
```
brandonbloom, line 766: same comment, use the pre-allocation & index setting.

brandonbloom (approved):
LGTM, all stylistic comments that should be quick to fix. Just helping to level-up your Go-game :)

internal/core/server/workspace.go
BenElgar: My understanding is that that should be impossible

PR #131, opened 2021-08-10
Overhaul Layout to use a single compact left navigation bar

jwmza:
This is the first in a series of GUI changes to move to the new GUI design, what I would call a _"true V1"_ of the UI as the previous one was never optimized for the things we care about in **exo**.

The new UI will (a) improve information density and reduce whitespace across the application while maintaining good aesthetic and legibility levels, (b) get rid of the three-side header+navbar+footer layout and replace it with a single left navbar layout to increase vertical space for use cases like log viewing and filtering, (c) adopt better underlying practices in UI engineering and use things like theme variables and user configuration to make the interface better suited to any user.

Here are some screenshots from the current state of this PR:

image
image

Here is a rough mockup of the intended direction for the UI to change in:

image

(Note above, in addition to the current changes, improvements to layout, iconography, and information density in the actual content panels)


gui/src/components/WorkspaceNav.svelte
```diff
@@ -0,0 +1,82 @@
...
+  export let workspaceId: string;
+  export let active: string;
+</script>
+
+<aside>
+  <button
+    on:click={() => {
+      router.push(`#/workspaces/${workspaceId}`);
+    }}
+    class={active === 'Dashboard' ? 'active' : ''}
```
brandonbloom, line 18: There is a nice shorthand for this pattern!

I think it is this:

```
    class:active={active === 'Dashboard'}
```

See: https://svelte.dev/docs#class_name

gui/src/components/WorkspaceNav.svelte
```diff
@@ -0,0 +1,82 @@
...
+    class={active === 'Storage' ? 'active' : ''}
+    title="Storage"><Storage /></button
+  >
+  <button
+    on:click={() => {
+      router.push(`#/workspaces/${workspaceId}/networking`);
+    }}
+    class={active === 'Networking' ? 'active' : ''}
+    title="Networking"><Networking /></button
+  >
```
brandonbloom, line 41: is this how prettier formats these? yuck.

brandonbloom (approved):

gui/src/components/WorkspaceNav.svelte
jwmza: Yes, but it can be fixed, which I will do.

jwmza:
@brandonbloom I think this one is ready to merge now.

Since you approved, there have been two significant changes:

- Everything in this PR now uses theme variables for color styling. This is an approach while in general should reduce the clutter/bloat of our styling code and will allow us to much more easily adopt dark mode or any other theme config later on. Right now it only affects the files affected by the PR itself, but it marks the beginning of a total switchover to CSS variables instead of in-place color definitions.
- I've removed the "back button" from `Layout`. As shown below, I think we could put it in the top-left-most panel of whatever page the user is on if we really want it, but given that browser windows already tend to have it and the global nav allows the user to navigate through known routes, I'm not sure it's even worth bringing back. Regardless, it needed to go from the left navbar. Too much unnecessary complication to global nav.
image

brandonbloom:
Still LGTM

PR #133, opened 2021-08-10
First pass at jobs gui

brandonbloom:
Joint work with @kendru - should be safe for main branch, but needs lots of follow-up work.

### DIFF
--- a/gui/src/components/Link.svelte
+++ b/gui/src/components/Link.svelte
@@ -1,0 +1,15 @@
+<script lang="ts">
+  import * as router from 'svelte-spa-router';
+
+  export let href: string;
+</script>
+
+<a
+  {href}
+  on:click={(event) => {
+    if (href.startsWith('#')) {
+      router.push(href);
+      event.preventDefault();
+    }
+  }}><slot /></a
+>
--- a/gui/src/pages/Job.svelte
+++ b/gui/src/pages/Job.svelte
@@ -1,0 +1,22 @@
+<script lang="ts">
+  import Layout from '../components/Layout.svelte';
+  import JobTree from '../components/JobTree.svelte';
+  import { api } from '../lib/api';
+
+  export let params = { job: '' };
+
+  const jobId = params.job;
+
+  const tasks = api.kernel.describeTasks({ jobIds: [jobId] });
+</script>
+
+<Layout showBackButton backButtonRoute="#/jobs">
+  <!-- TODO: common loader component -->
+  {#await tasks}
+    Loading...
+  {:then tasks}
+    <JobTree {jobId} {tasks} />
+  {:catch}
+    Error
+  {/await}
+</Layout>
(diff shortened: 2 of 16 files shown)


brandonbloom:
@jwmza look this over please - then let's chat soon about what we want to do with this

jwmza:
Will do

jwmza:
Well, this will definitely require adapting to the new layout and table component changes.

Even in the current version:

- The pages need a white background, and probably some padding too. I've created a new `MonoPanel` base layout for generic pages in the branch `left-nav-layout-overhaul`.
- Probably a good idea to use the `CheckeredTableWrapper` component around the tables here and use `tbody` in addition to `thead` to segment the table DOM tree properly and activate the generic checkered table styling.

I'm not sure how to test the actual jobs feature yet, will need some filling in on that @brandonbloom

gui/src/components/Link.svelte
```diff
@@ -0,0 +1,15 @@
+<script lang="ts">
+  import * as router from 'svelte-spa-router';
+
+  export let href: string;
+</script>
+
+<a
```
jwmza, line 7: Looks good. Might need an option to `display: block` as we use this broadly. Not sure yet.

gui/src/pages/Job.svelte
```diff
@@ -0,0 +1,22 @@
...
+
+  export let params = { job: '' };
+
+  const jobId = params.job;
+
+  const tasks = api.kernel.describeTasks({ jobIds: [jobId] });
+</script>
+
+<Layout showBackButton backButtonRoute="#/jobs">
+  <!-- TODO: common loader component -->
```
jwmza, line 14: Maybe wrap everything within `Layout` here inside a `<section>...</section>` that has `min-height: 100vh;` and `background: #ffffff` or wait until #131 lands and use `<MonoPanel>...</MonoPanel>`

gui/src/pages/Jobs.svelte
```diff
@@ -0,0 +1,57 @@
...
+  <section>
+    Jobs
+    <!-- TODO: common loader component -->
+    {#await jobs}
+      Loading...
+    {:then jobs}
+      {#if jobs.length === 0}
+        <i>None</i>
+      {:else}
+        <table>
```
jwmza, line 22: Suggestion: Wrap in `<CheckeredTableWrapper>` and use `<tbody>` in addition to `<thead>`

jwmza (commented):
LGTM, But I haven't tested the Tree stuff yet.

gui/src/pages/Job.svelte
brandonbloom: Going to wait for #131 - this UI is expectedly ugly as sin still.

brandonbloom:
merged in to the destroy-and-apply-progress branch

PR #134, opened 2021-08-10
Export manifests

kendru:
This adds initial support for exporting unix processes to a Procfile.


cmd/compose2exo/main.go
```diff
@@ -4,7 +4,7 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/deref/exo/internal/import/compose"
+	"github.com/deref/exo/internal/external/compose"
```
brandonbloom, line 7: `/internal/external/` is kinda confusing :-P Alternative name suggestion: `compat`? Or maybe just move `compose` and `procfile` in to the `manifest` package as sub-packages?

internal/core/api/workspace.josh.hcl
```diff
@@ -185,6 +204,7 @@ struct "process-description" {
   field "id" "string" {}
   field "provider" "string" {}
   field "name" "string" {}
+  field "spec" "interface{}" {}
```
brandonbloom, line 207: Is the intention here that this should be of type `process.Spec` from the `providers/unix/components/process` package? Maybe needs a comment, as this is a shortcoming of our code generator now.

internal/core/server/workspace.go
```diff
@@ -624,15 +630,22 @@ func (ws *Workspace) DescribeProcesses(ctx context.Context, input *api.DescribeP
 		case "process":
 			var state process.State
 			if err := jsonutil.UnmarshalString(component.State, &state); err != nil {
-				// TODO: log error.
-				fmt.Printf("unmarshalling process state: %v\n", err)
+				logger.Infof("unmarshalling process state: %v\n", err)
```
brandonbloom, line 633: thanks for fixing this :)

internal/core/server/workspace.go
```diff
@@ -747,6 +760,84 @@ func (ws *Workspace) DescribeNetworks(ctx context.Context, input *api.DescribeNe
...
+}
+
+func (ws *Workspace) resolveWorkspacePath(ctx context.Context, relativePath string) (string, error) {
+	description, err := ws.describe(ctx)
+	if err != nil {
+		return "", fmt.Errorf("describing workspace: %w", err)
+	}
+
+	resolvedPath := path.Join(description.Root, relativePath)
+	if !pathutil.HasPathPrefix(resolvedPath, description.Root) {
```
brandonbloom, line 834: glad you found this :)

internal/core/server/workspace.go
```diff
@@ -747,6 +760,84 @@ func (ws *Workspace) DescribeNetworks(ctx context.Context, input *api.DescribeNe
...
+
+func (ws *Workspace) ReadFile(ctx context.Context, input *api.ReadFileInput) (*api.ReadFileOutput, error) {
+	resolvedPath, err := ws.resolveWorkspacePath(ctx, input.Path)
+	if err != nil {
+		return nil, err
+	}
+
+	content, err := os.ReadFile(resolvedPath)
+	if err != nil {
+		return nil, fmt.Errorf("reading file: %w", err)
```
brandonbloom, line 801: probably don't want the `Errorf` here, as this is really the main crux of the `ReadFile` function & so doesn't add real context. Callers might wind up with `reading file: reading file: ...`

internal/core/server/workspace.go
```diff
@@ -747,6 +760,84 @@ func (ws *Workspace) DescribeNetworks(ctx context.Context, input *api.DescribeNe
...
+		return nil, err
+	}
+
+	mode := os.FileMode(0644)
+	if input.Mode != nil {
+		mode = os.FileMode(*input.Mode)
+	}
+
+	if err := os.WriteFile(resolvedPath, []byte(input.Content), mode); err != nil {
+		return nil, fmt.Errorf("writing file: %w", err)
```
brandonbloom, line 821: same feedback as above

internal/external/procfile/generate.go
```diff
@@ -0,0 +1,48 @@
+package procfile
+
+import (
+	"bytes"
+
+	"gopkg.in/alessio/shellescape.v1"
+	"mvdan.cc/sh/v3/syntax"
+)
+
+func Generate(procs []Process) (string, error) {
```
brandonbloom, line 10: Preferable to operate on an io.Writer here, as that's the most general thing:

```suggestion
func Generate(w io.Writer, procs []Process) error {
```

internal/external/procfile/generate.go
```diff
@@ -0,0 +1,48 @@
...
+		out.WriteByte('\n')
+	}
+
+	return out.String(), nil
+}
+
+func literalWord(val string) *syntax.Word {
+	return &syntax.Word{
+		Parts: []syntax.WordPart{
+			&syntax.Lit{Value: shellescape.Quote(val)},
```
brandonbloom, line 45: Does this work correctly? Is it _expected_ to work correctly? Seems like it might be abusing the fact that the printer doesn't actually inspect the internals of the Lit value... which might be OK? Not sure what else to do about it. Looks like you've got a test though, so works for me!

gui/src/components/ProcessList.svelte
```diff
@@ -160,6 +170,14 @@
       Error fetching process list: {processList.message}
     {/if}
   </div>
+  <div>
+    {#if procfileExport}
+      <p>
+        Your Procfile is not up to date.
+        <button on:click={() => workspace.writeFile('Procfile', procfileExport).then(checkProcfile)}>Export?</button>
```
brandonbloom, line 177: use async/await :)

gui/src/components/ProcessList.svelte
```diff
@@ -61,14 +61,24 @@
     loadInitialLogs(workspaceId, workspace);
   }
 
+  let procfileExport: string;
```
brandonbloom, line 64: If you use:

```suggestion
let procfileExport: string | null = null
```

Then below, you won't need the `undefined as any` cast.

internal/core/api/workspace.josh.hcl
kendru: The concrete type of this is going to depend on the component type. For a unix process, it is a `process.Spec`, but for a docker container, it would be a different type. Deserialization can happen in each branch of the switch where we build components from the serialized data.

internal/external/procfile/generate.go
kendru: Yeah, it looks like the parser generates `Lit` rather than more specific nodes indicating things like double quoted string, escaped character, etc. The shellescape library generates ugly output, but it is correct.

internal/core/api/workspace.josh.hcl
brandonbloom: In that case, I think this is buggy for Docker use cases, as we're going to serialize the JOSH, which will call MarshalJSON but really we want MarshalYAML. The type of the spec should just be `string`.

kendru: Okay. That kind of sucks that we have to unmarshal on use rather than on deserialization, but that makes sense. I'll change it and unmarshal it where I need it.

brandonbloom: Yeah, I've had some internal debate with myself on this topic. Ultimately, I decided that using a text format for the spec was the right thing to do because it lets us use shell scripts, docker compose yaml, terraform or hcl, etc directly without having to invent our own serialization formats. This of course does mean that any editing code paths or whatever will need to serialize/deserialize and lack generic editing of properties etc, but I don't think it is realistic for us to do generic editing of things like docker yaml, k8s yaml, etc which we don't control.

In general, treating it as an opaque string and marshalling/unmarshalling is maximally general, with all the pros and cons that usually brings along. For now, let's stick with that, and once we have more experience with an increasing number of components, we can decide if we want to invent a spec format that gives us the flexibility we need. For instance, we may want to use HCL so that we can have expressions or similar in there when we get to dynamically reconfigurable service constellations and need to include/exclude/change environment variables depending on what's running on host, docker, or cloud.

PR #136, opened 2021-08-10
Colors in logs

### LINKED ISSUE #123: [FEATURE] Support colourisation in logs
This applies both to the UI and the CLI.

brandonbloom:
Fixes #123 

This adds basic virtual terminal font attribute parsing logic to the FormattedLogMessage component.

See in Storybook:

image
image
image
image
image

And in the app:
image

### DIFF
--- a/gui/src/components/logs/FormattedLogMessage.svelte
+++ b/gui/src/components/logs/FormattedLogMessage.svelte
@@ -49,3 +49,10 @@
     {/if}
   {/each}
 </span>
+
+<style>
+  a,
+  span {
+    white-space: pre;
+  }
+</style>
--- a/gui/src/components/logs/FormattedLogMessage.svelte
+++ b/gui/src/components/logs/FormattedLogMessage.svelte
@@ -10,8 +10,8 @@
     const addStyle = (key: string, value: string) => {
       styles.push(`${key}: ${value}`);
     };
-    if (span.forground != null) {
-      addStyle('color', span.forground);
+    if (span.foreground != null) {
+      addStyle('color', span.foreground);
     }
     if (span.background != null) {
       addStyle('background', span.background);
--- a/gui/src/components/logs/FormattedLogMessage.svelte
+++ b/gui/src/components/logs/FormattedLogMessage.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import { parseSpans } from './parseSpans';
+  import { parseSpans, Span } from './parseSpans';
 
   export let message: string = '';
 
@@ -8,7 +8,7 @@
 
 <span>
   {#each spans as span}
-    {#if span.type === 'link'}
+    {#if span.href != null}
       <a href={span.href}>{span.text}</a>
     {:else}
       <span>{span.text}</span>
--- a/gui/src/components/logs/FormattedLogMessage.svelte
+++ b/gui/src/components/logs/FormattedLogMessage.svelte
@@ -5,8 +5,9 @@
 
   const spans = parseSpans(message);
 
-  const spanStyle = (span: Span): string => {
+  const spanProps = (span: Span): Record<string, string | boolean> => {
     const styles: string[] = [];
+    const classes: string[] = [];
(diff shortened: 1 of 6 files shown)


gui/src/components/logs/FormattedLogMessage.svelte
```diff
@@ -1,20 +1,51 @@
...
 
   const spans = parseSpans(message);
+
+  const spanStyle = (span: Span): string => {
+    const styles: string[] = [];
+    const addStyle = (key: string, value: string) => {
+      styles.push(`${key}: ${value}`);
+    };
+    if (span.forground != null) {
+      addStyle('color', span.forground);
```
jwmza, line 14: Nit: "foreground" instead of "forground"

gui/src/components/logs/FormattedLogMessage.svelte
```diff
@@ -1,20 +1,51 @@
...
+        break;
+      case 'faint':
+        addStyle('font-weight', 'lighter');
+        break;
+      case 'italic':
+        addStyle('font-style', 'italic');
+        break;
+      case 'underline':
+        addStyle('text-decoration', 'underline');
+      case 'blink':
```
jwmza, line 31: What's the intent here?

jwmza:
This is amazing! Does it already work with programs (e.g. the npm case we discussed yesterday) or do we still need to convince them to send the logs with color because we aren't a TTY?

gui/src/components/logs/FormattedLogMessage.svelte
brandonbloom: d'oh - will fix

gui/src/components/logs/FormattedLogMessage.svelte
brandonbloom: I think we'd have to do like `addClass('blink')`. I'll ping you on Slack & we can hack it up quick.

brandonbloom:
> do we still need to convince them to send the logs with color because we aren't a TTY?

Depends on the program, but often - yes - programs won't show colors if not in a TTY.

gui/src/components/logs/FormattedLogMessage.svelte
brandonbloom: fixed in 3152906

brandonbloom:
merging as there is test coverage, so feeling confident in it - but @BenElgar please do review when you get some time

gui/src/components/logs/FormattedLogMessage.svelte
```diff
@@ -1,20 +1,97 @@
 <script lang="ts">
-  import { parseSpans } from './parseSpans';
+  import { parseSpans, Span } from './parseSpans';
```
BenElgar, line 2: Doesn't work for me unless I import `Span` as a type.

gui/src/components/logs/parseSpans.ts
```diff
@@ -1,44 +1,239 @@
...
+    case 5:
+      return lookupColor256(codes[2]);
+    case 2:
+      return lookupColor24bit(codes);
+    default:
+      return '000000';
+  }
+};
 
 export const parseSpans = (input: string): Span[] => {
```
BenElgar, line 109: Not convinced by the name since it sounds like the function is parsing spans. I'd prefer `parseLogLine` or `logLineToSpans` or something.

BenElgar (commented):
Seems good to me :+1:

gui/src/components/logs/FormattedLogMessage.svelte
brandonbloom: fixed on main

gui/src/components/logs/parseSpans.ts
brandonbloom: Huh, I guess I never thought about the distinction between `parse${source}` and `parse${grammarProduction}` in naming conventions, but upon reflection, I've definitely seen both of these. The latter is much more common IME, so going to leave this alone.

PR #137, opened 2021-08-11
bump max log message length to 1 MiB

### LINKED ISSUE #126: [BUG] Logs are truncated
**Describe the bug**
Long log lines are truncated and/or split across lines.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a docker-compose.yaml like so:
```docker-compose
services:
  web:
    image: bash
    command: "printf '.%.0s' {1..9000}"
```
2. Run it with `exo run docker-compose.yaml`

**Expected behaviour**
One log line printed with 9000 dots on one line in both the UI and the CLI. Instead what happens is that some smaller number of dots is printed in the UI (this seems to be consistently 8080) and none are printed in the CLI.

If the 9000 is increased to a much larger number the dots appear to get split across multiple lines in the UI but still don't sum to the correct number.

brandonbloom:
Fixes #126

Though will probably introduce a follow-up issue that long log lines perform badly :-P

### DIFF
--- a/internal/logd/api/consts.go
+++ b/internal/logd/api/consts.go
@@ -1,3 +1,6 @@
 package api
 
-const MaxMessageSize = 1024 // Includes space for newline terminator.
+// Max length of log messages. Size chosen to fit within a UDP packet (max size
+// 64k) and reserve space for Syslog event headers. The message itself always
+// includes a newline terminator as well.
+const MaxMessageSize = 48 * 1024


brandonbloom:
@kendru I think that now that we use syslog and badger, this is safe to do - but maybe there is a max value size or something somewhere else I couldn't find? Not sure how to validate this other than to do some performance testing with large messages.

brandonbloom:
@kendru if you think this is ok, please go ahead and merge whenever :) thanks

kendru:
@brandonbloom as long as the syslog client stuff can split it into multiple packets, then that should be good. UDP has a max packet size of ~45Kb, but there are a lot of reasons that the effective max size is smaller.

PR #139, opened 2021-08-11
Make exo run tail logs

BenElgar:


cmd/exo/run.go
```diff
@@ -68,8 +68,8 @@ If a workspace does not exist, one will be created in the current directory.
 		(func() {
 			ctx, stop := signal.NotifyContext(ctx, os.Interrupt, syscall.SIGTERM)
 			defer stop()
-			var logRefs []string
-			if err := tailLogs(ctx, workspace, logRefs); err != nil {
+			refs := []string{}
```
brandonbloom, line 71: This was intentionally nil before, to signal "no logs specified" differently than "specified no logs". I think you can leave this point alone and change the `len(logRefs) != 0` to `logRefs != nil`

BenElgar: Yeah this is leftover from my first attempt at fixing. Will revert.

BenElgar (commented):
Okay yesterday when I made this PR I was consistently not getting any logs from `exo run` and indeed hadn't ever. `logIDs` was always an empty slice and `GetEvents` always returned no events. I thought that this was the fix but I can now no longer reproduce on master. :man_shrugging: 

I'm going to close and reopen/reinvestigate if I see it again.

PR #141, opened 2021-08-11
Add support for command in docker image

BenElgar:


internal/providers/docker/compose/yaml.go
```diff
@@ -52,7 +52,9 @@ type Service struct {
 	// TODO: cap_add
 	// TODO: cap_drop
 	// TODO: cgroup_parent
-	Command       Command  `yaml:"command"`
+
+	Command Command `yaml:"command"`
+
```
brandonbloom, line 57: what's with the extra whitespace here?

internal/providers/docker/components/container/image.go
```diff
@@ -29,6 +29,12 @@ func (c *Container) ensureImage(ctx context.Context) error {
 		return fmt.Errorf("inspecting image: %w", err)
 	}
 	c.ImageID = inspection.ID
+
+	c.Command = make([]string, len(inspection.Config.Cmd))
+	for i, cmd := range inspection.Config.Cmd {
+		c.Command[i] = cmd
+	}
```
brandonbloom, line 36: Two problems with this:

1) Mutating the spec, not the state. Subsequent controller invocations won't have this information, so a `start` or `restart` operation (which doesn't call `ensureImage`) will not have access to the correct command.

2) This will shadow the docker compose file's command, which is the opposite of what we want, right? Presumably the compose file should shadow the code in the image.

brandonbloom (changes requested):
See comments. I think we want to put more/all of the image inspection result in to the state & then have some sort of `EffectiveCommand` state as well.

internal/providers/docker/compose/yaml.go
BenElgar: The go commenting convention is to have a comment directly preceding the thing you're documenting. This means that tools expect any comment directly above a struct member to be documentation for that struct member. So the documentation for `Command`, as reported by `go doc` and IDEs/language servers, was "TODO: cap_add TODO: cap_drop".

But the whitespace below this line was unnecessary. Will remove.

internal/providers/docker/components/container/image.go
BenElgar: This is perhaps an aside but shouldn't both `start` and `restart`[0] call `ensureImage`? There's no guarantee the image will still be around if the container isn't running any more.

BenElgar:
Addressed the above

internal/providers/docker/components/container/lifecycle.go
```diff
@@ -42,6 +42,22 @@ func (c *Container) create(ctx context.Context) error {
...
+	entrypoint := strslice.StrSlice(c.Spec.Entrypoint)
+	if len(cmd) == 0 {
+		entrypoint = c.State.Image.Entrypoint
+	}
+
+	workingDir := c.Spec.WorkingDir
+	if workingDir == "" {
+		workingDir = c.State.Image.WorkingDir
+	}
+
```
brandonbloom, line 60: super nit-picky, but here you are creating some locals and then assigning them in the `container.Config{` literal. However, right below that, we've got a bunch of code that is assigning directly in to the `containerCfg` object. The way you did it means that you can see a full list of assigned properties in one literal, but it also means having to name temporaries all over the place. Either way is fine, but it would be nice to follow the pattern.

```go
containerCfg.WorkingDir = c.Spec.WorkingDir
if containerCfg.WorkignDir == "" {
  containerCfg.WorkingDir = c.State.Image.WorkingDir
}
```

internal/providers/docker/components/container/lifecycle.go
```diff
@@ -51,13 +67,14 @@ func (c *Container) create(ctx context.Context) error {
...
 		Env:         c.Spec.Environment.Slice(),
-		Cmd:         strslice.StrSlice(c.Spec.Command),
+		Cmd:         cmd,
 		Healthcheck: healthCfg,
 		// ArgsEscaped     bool                `json:",omitempty"` // True if command is already escaped (meaning treat as a command line) (Windows specific).
-		Image: c.State.ImageID,
+
+		Image: c.State.Image.ID,
 		// Volumes         map[string]struct{} // List of volumes (mounts) used for the container
 		WorkingDir: c.Spec.WorkingDir,
```
brandonbloom, line 76: Did you intend to set `WorkingDir: workingDir,` here?

internal/providers/docker/components/container/component.go
```diff
@@ -17,7 +18,14 @@ type Container struct {
...
 type State struct {
-	ImageID     string `json:"imageId"`
-	ContainerID string `json:"containerId"`
-	Running     bool   `json:"running"`
+	ContainerID string     `json:"containerId"`
+	Running     bool       `json:"running"`
+	Image       ImageProps `json:"image"`
+}
+
+type ImageProps struct {
```
brandonbloom, line 26: I've been intentionally avoiding the word `Props` in these names for two reasons: 1) in react/ui world props ~ specs and 2) in terraform, pulumi, etc, props = inputs + outputs.

A better name here might be `ImageState`?

brandonbloom (approved):
LGTM, except the the workingDir value not making it on to containerCfg.

brandonbloom:
Unrelated to anything else: Please rebase this. There are 9 commits in here, 3 of which are merge commits. As I try to go through the history to build a change log, I find this noise pretty hard to parse.

PR #142, opened 2021-08-11
Improve process list styling

jwmza:
image

https://loom.com/share/8a440b7cbb054d949aac49caa65bdf4a

### DIFF
--- a/gui/public/globals.css
+++ b/gui/public/globals.css
@@ -4,7 +4,7 @@
 :root {
   font-family: Inter, -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto,
     Oxygen, Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;
-  color: var(--text-color);
+  color: var(--color);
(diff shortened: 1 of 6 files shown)


brandonbloom (approved):
This LGTM, but a few thoughts:

1) I'm a little bit nervous about how the css variables appear to be globally scoped names. Not sure if there is a modularity mechanism there for that, but I see you've at least prefixed names in a reasonable way.

2) The `ProcessList` component is getting large! My advice would be to create a directory for it and it's subcomponents and start to break it up in to smaller and smaller pieces. Smaller components are easier to reason about and test. Especially with storybook.

jwmza:
@brandonbloom CSS variables can be overwritten and cascade down to DOM children - the theme variables *should* be globally scoped though, because they are what is going to allow us to have dark mode and perhaps some day even user customizable themes with a simple single file system (this is roughly what every major theme-able application does)

Wouldn't be a bad idea to get into the habit of namespacing these systematically when they concern modular components, though.

jwmza:
Agreed re: smaller components

PR #143, opened 2021-08-11
Stop exo run when process terminates

BenElgar:


cmd/exo/run.go
```diff
@@ -68,8 +68,9 @@ If a workspace does not exist, one will be created in the current directory.
 		(func() {
 			ctx, stop := signal.NotifyContext(ctx, os.Interrupt, syscall.SIGTERM)
 			defer stop()
-			var logRefs []string
-			if err := tailLogs(ctx, workspace, logRefs); err != nil {
+			refs := []string{}
```
brandonbloom, line 71: See comment here: https://github.com/deref/exo/pull/139#issuecomment-896971116

cmd/exo/logs.go
```diff
@@ -109,6 +113,22 @@ func tailLogs(ctx context.Context, workspace api.Workspace, logRefs []string) er
...
+		if stopOnError {
+			descriptions, err = workspace.DescribeProcesses(ctx, &api.DescribeProcessesInput{})
+			if err != nil {
+				return fmt.Errorf("failed to check status of processes: %w", err)
+			}
+
+			for _, proc := range descriptions.Processes {
+				for _, id := range logIDs {
+					if proc.ID == id && !proc.Running {
+						return fmt.Errorf("process stopped running")
```
brandonbloom, line 126: nit: use `errors.New` when there are no dynamic formatting bits in the error message. Though, here, I think it would make sense to say _which_ process stopped running. Ideally, the component name. Something like:

```suggestion
						return fmt.Errorf("process stopped running: %q", proc.Name)
```

brandonbloom (commented):
The stopOnError stuff looks reasonable to me, but I'm concerned about the stuff inherited from #139

PR #150, opened 2021-08-11
Fix telemetry

kendru:
The telemetry changes caused exo to silently fail, so the merge was reverted. This PR contains the telemetry code and will soon have the fix applied.

### DIFF
--- a/internal/telemetry/default.go
+++ b/internal/telemetry/default.go
@@ -36,7 +36,9 @@ func (t *defaultTelemetry) IsEnabled() bool {
 }
 
 func (t *defaultTelemetry) StartSession(ctx context.Context) {
-	t.ensureSession(ctx)
+	go func() {
+		t.ensureSession(ctx)
+	}()
 }
 
 func (t *defaultTelemetry) LatestVersion(ctx context.Context) (string, error) {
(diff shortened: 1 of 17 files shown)


kendru:
@brandonbloom I ran with this branch yesterday without any issues. Could you do a sanity check again to see if it is still busted for you?

brandonbloom:
I just did:

```
git fetch --all
git checkout -b fix-telemetry origin/fix-telemetry
git merge origin/main
./script/install-release.sh
exo daemon
```

And got as output: `daemon not healthy`

kendru:
@brandonbloom The problem was that the server was not starting fast enough because it was trying to start the session synchronously. I just moved that into a go routine, since any call to send an event will not be scheduled until the session is actually initialized.

internal/telemetry/default.go
```diff
@@ -36,7 +36,9 @@ func (t *defaultTelemetry) IsEnabled() bool {
 }
 
 func (t *defaultTelemetry) StartSession(ctx context.Context) {
-	t.ensureSession(ctx)
+	go func() {
+		t.ensureSession(ctx)
+	}()
```
brandonbloom, line 41: this is just `go t.ensureSession(ctx)`

PR #152, opened 2021-08-11
Defer fetching logs until visible logs are present.

### LINKED ISSUE #125: [BUG] Log pages load very slowly in wrong order in the UI
If you view a process that has lots of logs it will load the logs forwards from the first line. This can be very slow if there are many thousands of lines of logs. Instead I'd propose that we load the most recent logs and then page backwards.

kendru:
Fixes #125


doc/local-development.md
```diff
@@ -36,3 +36,15 @@ working on the project.
 To test CLI changes against the auto-restarting watched development server, use
 `./script/dexo`. If you're using direnv (see above), then `dexo` will be on
 your `PATH`.
+
+### Troubleshooting
+
+#### The GUI won't start. I see an "Error: esbuild: Failed to install correctly" error in the log.
```
brandonbloom, line 42: What causes this?

brandonbloom:
I'm curious about the mechanism here. Does this mean that you won't get any scrollback when you launch the GUI in the future?

Seems less than ideal. Can we make the initial "since" like -10 minutes or something to give you at least some scrollback?

PR #161, opened 2021-08-12
Improve UI density and code

jwmza:
- [x] Improves `TextEditor` styling
- [x] Cleans up `NewProcess` layout and styling
- [x] Implements `MonoPanel` with improved styling and optional padding across all relevant pages
- [x] Cleans up repetitive and redundant page-level code for layout and spacing
- [x] Clean up repetitive panel-title heading styling
- [x] Create and use `TwoPanel` component for e.g. workspace page
- [ ] Extract excessive page-level styling and components to `/components`


gui/src/components/Panel.svelte
```diff
@@ -0,0 +1,80 @@
+<script lang="ts">
+  import IconButton from './IconButton.svelte';
+  import BackSVG from './mono/leftarrow.svelte';
+  import * as router from 'svelte-spa-router';
+
+  export let title: string | undefined = undefined;
```
brandonbloom, line 6: There is a shorthand for this:

```typescript
  export let title: string;
```

However, in this case, I think you want:

```typescript
export let title: string | null = null;
```

The reason being that you generally want to avoid using `undefined` as a value. In this case in particular, the goal is to convey "the title is set to none", not "the title has not yet been set".

gui/src/components/Panel.svelte
```diff
@@ -0,0 +1,80 @@
...
+  export let backRoute: string | undefined = undefined;
+</script>
+
+<div class="panel">
+  {#if $$slots.actions || title !== undefined || backRoute !== undefined}
```
brandonbloom, line 11: If you take the above advice, this becomes `title !== null`. However, a generally more robust check might be `title != null` which checks for == null || == undefined.

Also, while I generally prefer explicit boolean checks rather than relying on "truthiness", it's actually really common and convenient in UI code. So you could even just shorten this to `|| title || backRoute` and drop the null/undefined checks altogether. In that case, you could also change the export to `export let title: string = '';`

gui/src/components/Panel.svelte
```diff
@@ -0,0 +1,80 @@
...
+    <header>
+      <div class="header-title">
+        {#if backRoute !== undefined}
```
brandonbloom, line 14: Again, in template code, really convenient to just write `{#if backRoute}` even tho it's less explicit, it's basically _why_ JavaScript has this sort of truthiness checks in the first place.

gui/src/pages/NewProcess.svelte
```diff
@@ -48,11 +48,21 @@
       arguments: args,
     });
   };
+
+  const codeExample = `# Export environment variables.
```
brandonbloom, line 52: heh, why didn't i think of this? :-P

brandonbloom (approved):
LGTM besides some nitpicks.

jwmza:
Thanks @brandonbloom will do

jwmza:
Also now wraps log messages 

image

brandonbloom:
@jwmza you're starting to see some merge conflicts... getter to get smaller changes to the main branch as you go

jwmza:
Yeah I'm moving to merge this ASAP @brandonbloom

jwmza:
Will follow up with componentization from complex page UIs later.

PR #164, opened 2021-08-12
Make dexo clean up after itself

BenElgar:


script/dexo
```diff
@@ -10,7 +10,9 @@ script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
 
 pushd "$script_dir/.." > /dev/null
 
-tmp_dir=$(mktemp -d)
+tmp_dir=$(mktemp -d -t tmp-dexo-XXXXXXX)
+trap 'rm -rf -- "$tmp_dir"' EXIT
```
brandonbloom, line 14: Thanks @BenElgar, but I can't help but wonder:

@kendru why use `go build` instead of `go run`? In my experience (and some testing locally just now) `go run` is dramatically faster than `go build` and cleans itself up too. For a trivial change in `./cmd/exo`, I'm seeing build take 3.5 seconds and run take 1.25 seconds.

brandonbloom: D'oh. Gotcha.

Maybe we should add a `-C` global flag, like `make` has? So:

```bash
wd=$PWD
cd $EXO_CODE_HOME_OR_WHATEVER
exo -C $wd ...
```

BenElgar: That works though I'm not sure I find the problem pressing enough to solve.

brandonbloom: I might do it, since the perf difference of build & run is a huge impact on my workflow.

brandonbloom: Also, `-C` is just generally useful - i already have `(cd ~/Projects/exo && exo restart server)` and similar in my bash history :)

PR #170, opened 2021-08-13
support "expose" and port ranges from compose files

brandonbloom:
I haven't tested this at all yet. I saw that you mentioned `expose` in our compatibility tracking spreadsheet, so I resurrected this code from a local branch and at least got it compiling.

@BenElgar would you prefer I finish it up or would you be willing to test this and carry it across the finish line?


BenElgar:
Working on it now

brandonbloom:
@BenElgar since I opened this PR, I can't "approve" it :) but it LGTM. Feel free to merge if you've tested it and it's working. Thanks!

internal/providers/docker/components/container/lifecycle.go
```diff
@@ -165,14 +187,41 @@ func (c *Container) create(ctx context.Context) error {
...
+				return fmt.Errorf("could not parse port range: %w", err)
+			}
+
+			publishedLow, publishedHigh, err := publishedPort.Range()
+			if err != nil {
+				return fmt.Errorf("could not parse port range: %w", err)
+			}
+
+			hostPort := publishedPort.Port()
+			if publishedHigh != publishedLow {
```
brandonbloom, line 212: Not sure I quite understand this bit. Also, I don't see any check to validate that the published/host ranges have the same size/length/number-of-ports... They are parallel arrays, right?

BenElgar: Yes you're right — it didn't break when it should have. I've added some more validation now.

PR #176, opened 2021-08-13
Fix job contexts

BenElgar:


BenElgar:
Not 100% sure this is what was intended but merging in order to fix master so progress can be made

PR #177, opened 2021-08-13
Clean exod shutdown

### LINKED ISSUE #148: [BUG] `dexo exit` does not exit the dev daemon
**Describe the bug**
When using the `dexo exit` command, it should exit the dev process. It currently exits the installed instance rather than the dev one.

**To Reproduce**
1. `cd` to the repo root directory
2. Run `exo run`, then press <kbd>Ctl</kbd>+<kbd>C</kbd> to stop the logs process.
3. Verify that the dev server is running: `dexo status`
4. Run `dexo exit`

**Expected behavior**
After exiting, `dexo status` should indicate that the dev server is no longer running.

kendru:
Now that the process will try to exit cleanly on sigterm, this moves the code for sending the server a sigterm from the client and into the server itself. `exo exit` then becomes an RPC call that instructs the server to try to terminate itself. This also changes a lot of the CLI commands to not attempt to start a daemon if a `client.url` is present in the config file.

Fixes #148

### DIFF
--- a/cmd/exo/daemon.go
+++ b/cmd/exo/daemon.go
@@ -39,6 +39,26 @@ var runState struct {
 	URL string `json:"url"`
 }
 
+// checkOrEnsureServer is intended to be used by the CLI commands that do not explicitly start a server.
+// It will first check whether a server is running, and if it is not, the behaviour depends on whether
+// a URL is configured for the exo client or not:
+// - If a URL is configured, this fails since the server is not available.
+// - If no URL is configured, this attempts to start a server and exits with an error if it cannot.
+// The assumption is that if a URL is configured, the client should not try to start a local instance. One
+// implication of this is that if you are running a local server on a non-default port, you will need to
+// explicitly run one of the following commands to explicitly start a server: `exo daemon`, `exo server`,
+// `exo run`.
+func checkOrEnsureServer() {
+	if checkHealthy() {
+		return
+	}
+	if cfg.Client.URL == "" {
+		ensureDaemon()
+		return
+	}
+	cmdutil.Fatalf("the server at %q is not healthy", cfg.Client.URL)
+}
+
 func ensureDaemon() {
 	// Validate exod process record.
 	err := loadRunState()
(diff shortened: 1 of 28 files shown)


internal/util/cmdutil/http.go
```diff
@@ -20,7 +20,7 @@ func ListenAndServe(ctx context.Context, svr *http.Server) {
 		ShutdownOnInterrupt(ctx, svr)
 		return nil
 	})
-	if err := eg.Wait(); err != nil {
+	if err := eg.Wait(); err != nil && err != http.ErrServerClosed {
```
kendru, line 23: This is the key thing that was preventing Badger from exiting cleanly.

internal/core/server/kernel.go
```diff
@@ -116,10 +117,24 @@ func (kern *Kernel) Ping(context.Context, *api.PingInput) (*api.PingOutput, erro
 	return &api.PingOutput{}, nil
 }
 
+func (kern *Kernel) Exit(context.Context, *api.ExitInput) (*api.ExitOutput, error) {
+	// Return immediately, then asynchronously try to shutdown.
+	defer func() {
+		go func() {
```
brandonbloom, line 123: What is the purpose of the `defer func() { go func() {` here? Seems like simply `go func() {` would do essentially the same thing. If the `go` was causing it to exit _before_ returning the response to a client, that's a race condition. Seems like the `defer` here is akin to `sleep`.

I suspect that the right thing to do here is to actually make `Exit` a non-josh method and instead make a custom HTTP route/handler for it. This way, the server can be sure to return a 200 OK to the client _before_ starting to exit.

brandonbloom (commented):
LGTM overall. Two concerns are:

1) killExod vs killing self
2) exit handler race condition

cmd/exo/daemon.go
```diff
@@ -92,7 +112,7 @@ func ensureDaemon() {
 	// Cleanup if unhealthy.
 	if !ok {
 		cmdutil.Warnf("daemon not healthy")
-		killExod()
+		osutil.TerminateProcessWithTimeout(os.Getpid(), time.Second*5)
```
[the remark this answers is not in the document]
kendru: This is my bad - will fix.

internal/core/server/kernel.go
kendru: Yes, this does essentially function the same as a sleep. I'm happy to make a separate handler for it, but right now the only means we have of shutting down cleanly is externally through a process signal rather than some handle that is passed down into the `Kernel` struct. I'll remove the defer for now since it probably has no effect in most cases, and rely on the call to `svr.Shutdown(ctx)` flush this connection before shutting down.

PR #185, opened 2021-08-14
Add license command & embed fonts

### LINKED ISSUE #145: [BUG] Fonts not bundled with release
We're using fonts that are not bundled with the release, so if the user doesn't have them installed, they don't work!

brandonbloom:
@jwmza you can do your font embedding work off of this branch

Fixes #145


jwmza (approved):

jwmza:
I believe this is now good to go and fixes #145 

However, it should be noted that this includes **all** the WOFF files for both fonts, and we don't use the extreme/uncommon weights. It's around ~6 MB of font files. In theory if we want to be storage efficient we could remove some, but this doesn't seem high priority atm.

jwmza:
@brandonbloom please note: I've disabled ligatures by default for both logs and code as per discussion in the related issue.

We should certainly introduce this as user-preference config later on; however, I would strongly advocate for making ligatures default-**on** once there is a toggle to turn them off. I estimate that the net negatives of user preferences around ligatures for people who for whatever reason aren't aware or don't decide to use the manual toggle will tilt towards an overall preference for ligatures. (The ligature feature also controls some variable glyphs that aren't _technically_ ligatures, e.g. the vertical placement of disjointed dashes and colons/semicolons depending on their surrounding concept is greatly improved with ligatures on, despite the glyphs not being visibly connected.

brandonbloom:
On the topic of ligatures: Let's leave them disabled for now, and we can revisit later when we have a more robust preferences system. If there is a growing list of preferences, feel free to collect those somewhere so we can make sure ligatures makes the list.

On the topic of file size: There are two separate concerns. (A) deployed binary size and (B) git repository size.

First, a break down of our existing binary size. Right now, our release binary size is 27MiB, which is not tiny, but not huge either. Adding an extra 6MiB puts us at 33MiB. From what I can tell, about 9MiB of that is "runtime" data from Go, which is a known problem with Go binaries growing pretty quickly due to debug info such as the program counter to line numbers table. There is also 6.6MiB of "unknown" data, which I don't have an obvious tool for investigating.

I don't see any reason to worry about binary size yet, though it is something we should keep an eye on.

Regarding repository size: If you add these files, they will be in the git tree _forever_. And anyone who ever clones exo will get every copy of every font file we've ever used. This is a major shortcoming of the design of Git, but it is unavoidable. The only things you can really do about it are to either avoid adding lots of binary files, or eventually give up and split up the repo, rewriting history, or move to Git "large file support" or a variety of other awkward, complexity-adding hacks.

The main issue to be aware of regarding repository size is frequently changing large files. Git's delta compression can help, but it's still a problem. Inter in particular seems to release new versions of the font every few months, where as Firacode looks a little more stable. As long as we don't track the latest releases of those regularly, we're good I think.

I'll go ahead and merge this.

jwmza:
Thanks for the detailed explanation @brandonbloom 

I think we should be fine for now. I don't see a reason for concern around repository size, as I don't see a reason to frequently update fonts unless something major and unexpected about them changes - the most reasonable thing I can see happening within ~60 months is we might want to reduce the binary size by one or two MB by removing unused font files, but honestly I don't know if the binary even contains the unused ones to begin with so this might be a total non problem. Whatever is building the releases might already not include font files which aren't even imported into a used CSS file.

As for user preferences I'll make a doc

brandonbloom:
> I don't see a reason for concern around repository size

Because I have personally experienced a lack of concern for this turn in to a major problem surprisingly quickly on several other project's I've contributed to :) This is just a limitation of Git that you have to pay attention to.

> honestly I don't know if the binary even contains the unused ones to begin with so this might be a total non problem.

It absolutely does. Since arbitrary code can load arbitrary assets, dead code elimination for files in `public` is an undecidable problem, so essentially all JS build tools do the simplest possible thing: Just copy all the public files in to the build output directory. Then we embed the entire build output directory when we bundle the exo binary.

jwmza:
Re: `/public/`, yeah, that makes a lot of sense now that you put it that way. So, it is probably a good idea to remove the unnecessary font weight files at some point. For now, I think we should wait until our visual style (in this case regarding typography) is stable and well-liked by users, which also means waiting for some more traction and feedback.

Will probably end up being something like `>300` and `<700` for all weights.

PR #192, opened 2021-08-16
Add exo labels to docker

BenElgar:


brandonbloom (changes requested):
The approach of adding labels as an argument to initialize is problematic for several reasons.

1) This initialize method applies to all kinds of resources and providers, so there is no universal key format for labels. What if we were initializing an AWS resource with tags instead of a Docker resource with labels? AWS uses colons instead of dots in the separator for tag names.

2) The information passed in is redundant. The workspace id and component id values are already passed in to callers and stored in the `ComponentBase` structure.

3) The InitializeInput struct is unavailable in other code paths that might need to set labels on docker resources. For instance, today we have an `update` codepath from `exo apply`, but it just naively does a delete/recreate. Consider if we enabled it for optimization, where instead of `Initialize` we call `Update` to intelligently update, recreating only if necessary. That interface would also need to accept this `exo-labels` argument or otherwise this information would need to be stored in component state.

4) It's not necessary to spread the knowledge of labels around multiple places in the code. You can simply add a method like `func (c *DockerComponentBase) ExoLabels() map[string]string`.

BenElgar:
Workspace ID didn't appear to be unless I'm misunderstanding:
```go
type ComponentBase struct {
	ComponentID   string
	WorkspaceRoot string
	Logger        logging.Logger
	isDeleted     bool
}
```
My thinking was that that was an intentional decision to maintain the abstraction and that we should pass in an opaque set of labels rather than leak every abstraction that we want to label through to the creation machinery, especially if we want to add a lot more labels.

But I don't feel particularly strongly about it and am happy to add workspace ID to the component struct.

I agree that the label names could be different but my assumption was that the providers would map them appropriately.

brandonbloom:
> But I don't feel particularly strongly about it and am happy to add workspace ID to the component struct.

I do feel pretty strongly about it. I think we should go that route.

> maintain the abstraction and that we should pass in an opaque set of labels rather than leak every abstraction that we want to label through to the creation machinery

What abstraction is breaking or leaking?

> especially if we want to add a lot more labels.

We can solve this problem by making a method on the various BaseComponent types, which is how we reuse code across controllers now. If later, we decide we want to hoist labels higher-up in the system, we have an easy method to find-references on.

> I agree that the label names could be different but my assumption was that the providers would map them appropriately.

If we do that, we need to carefully consider how global labels naming scheme. For now, I'd rather we do the more specialized thing specifically for docker and only generalize/hoist as the need becomes apparent when we add other label-supporting providers.

BenElgar:
I've made the changes.

The abstraction I was trying to maintain was that of the "executor". My thinking was that if we want to start adding labels to the resources like requesting user, price, time of request, version of manifest or whether the resource was requested via the UI or CLI, we don't want to just dump all those things into `ComponentBase`. Instead we probably want an opaque map that different components of the system can add to as the request percolates through the pipeline. But it's not a big deal and we can solve that problem when we come to it.

brandonbloom (approved):

Issue #180, opened 2021-08-13
[BUG] if docker is not installed, we don't handle that gracefully


brandonbloom:
Docker client operations fail with connection refused or similar. We should detect the docker agent not running, and present useful UX.

docker-compose actually checks if Docker is installed, and if not, points users to various install links. We don't need to go that far, but if there is an issue, we may want to provide some trouble shooting steps, up-to-and-including linking to the [Get Docker](https://docs.docker.com/get-docker/) page.

We should also consider what happens if there are version incompatibilities.

BenElgar:
I actually thought the handling was pretty graceful:
```
❯ dexo run docker-compose.y*
GUI available at: http://localhost:4001/#/workspaces/n2pdtmkbe0pcykwzna4g5xz644
Job URL: http://localhost:4001/#/jobs/pxj2yfamhw54j9p1efkjsavqmg
applying delete "test" for replacement: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
⨯ └─ updating test delete "test" for replacement: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
Error: applying manifest: job failure: delete "test" for replacement: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
applying manifest: job failure: delete "test" for replacement: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

What would you prefer?

brandonbloom:
I didn't actually test this to see how we behaved, I was just analyzing the code in docker-compose and they have lots of handling for this situation. Ugly jobs error reporting aside (which is a separate issue we should fix), I think you're right that this error message is fine.

PR #193, opened 2021-08-16
Add basic guidance if docker isn't running

BenElgar:


brandonbloom:
Apply is far from the only place that we would expect docker failures. However, since #180 wasn't actually as bad as expected, let's not bother investing the time and energy yet to get this right.

Issue #188, opened 2021-08-16
[FEATURE] `kill` operation


brandonbloom:
We have `stop`, but it has a default timeout. We could add a customizable timeout, but it's sometimes useful to instant-kill a process to see how the system behaves, or if you know it is wedged.

Additionally, it may be useful to send arbitrary unix signals to processes.

See docker-compose's support here:

```
$ docker-compose kill --help
Force stop service containers.

Usage: kill [options] [--] [SERVICE...]

Options:
    -s SIGNAL         SIGNAL to send to the container.
                      Default signal is SIGKILL.
```

brandonbloom:
I'm going to implement this quickly because I'm messing with some example that uses SIGHUP and it would be convenient to have this.

brandonbloom:
implemented by #239

PR #194, opened 2021-08-16
Add stop/restart timeout

BenElgar:


brandonbloom (commented):
i'm not done reviewing yet - just had to run to the team meeting :)

internal/core/api/lifecycle.josh.hcl
```diff
@@ -1,10 +1,12 @@
...
-  
+
   method "initialize" {}
-  
+
   method "refresh" {}
-  
+
   method "dispose" {
+    input "stop-now" "bool" {}
```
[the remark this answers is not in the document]
BenElgar: Yes, sorry, this is leftover from when I misunderstood what "dispose" was.

cmd/exo/stop.go
```diff
@@ -7,8 +7,11 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var stopNow = false
+
 func init() {
 	rootCmd.AddCommand(stopCmd)
+	stopCmd.Flags().BoolVar(&stopNow, "stop-now", false, "Attempts to stop the process immediately")
```
brandonbloom, line 14: I think that we should have "timeout" instead of "stop-now". Could be expressed in milliseconds, or we can maybe use the duration parsing logic - either from docker compose that I ported, or the builtin logic in Go.

internal/core/api/workspace.josh.hcl
```diff
@@ -5,10 +5,12 @@ interface "process" {
     output "job-id" "string" {}
   }
   method "stop" {
+    input "stop-now" "bool" {}
```
brandonbloom, line 8: yeah, i think all of these should be an optional timeout.

brandonbloom (commented):
So I think this _partially_ addresses #188, but a full solution would support signaling. See `docker-compose kill --help`. I'll update #188.

What this PR has now is an incomplete implementation of the `--timeout` option. See `docker-compose stop --help` and `docker-compose restart --help`. I don't think that this functionality is high-enough priority to justify racing to land an incomplete implementation that only supports timeout=0.

cmd/exo/stop.go
```diff
@@ -7,8 +7,13 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var initTimeout uint = 0
```
brandonbloom, line 10: you don't need this `initTimeout` variable, since it is only used in the `.Flags()` call. That line can instead do `&timeoutSeconds` if `var timeout int` instead of `*int`.

internal/core/api/workspace.josh.hcl
```diff
@@ -5,10 +5,12 @@ interface "process" {
     output "job-id" "string" {}
   }
   method "stop" {
+    input "timeout-seconds" "*uint" {}
     output "job-id" "string" {}
   }
   # TODO: Optional method?
   method "restart" {
+    input "timeout-seconds" "*uint" {}
```
brandonbloom, line 13: Both the stop and stop-components methods are used by `exo stop`, but you added this for restart without changing `exo restart`.

brandonbloom (approved):
LGTM with 1 nitpick and one missing bit: `exo restart foo --timeout 1`

cmd/exo/stop.go
BenElgar: Ah but I need to pass a pointer to `ws.Stop` that is possibly `nil`. If `timeoutSeconds` is `uint` rather than `*uint` I can't do that, since you can't do `&timeoutSeconds = nil`.

internal/core/api/workspace.josh.hcl
BenElgar: :+1:

PR #196, opened 2021-08-16
Separate the Logs UI & styling from the LogsViewer stateful component

jwmza:


gui/src/components/logs/logTypes.ts
```diff
@@ -0,0 +1,12 @@
...
+  style: string;
+  time: {
+    short: string;
+    full: string;
+  };
+  name: string;
+  message: string;
+}
+
+export type Logs = Log[];
```
kendru, line 12: Generally, type aliases that are just `T[]` are not all that useful and imply that there is more to the type.

kendru (approved):
LGTM. I think this is a step in the right direction.

gui/src/components/logs/logTypes.ts
jwmza: Fixed, thanks :)

PR #198, opened 2021-08-16
Remote data component

kendru:
This creates a RemoteData component that uses named slots to deal with data in various states of a `RequestLifecycle`. It also refactors the components that branched based on the `remoteData.stage` value to use Remote Data. The goal here is to start lifting the data handling code out of the components with display logic.

### DIFF
--- a/gui/src/components/RemoteData.svelte
+++ b/gui/src/components/RemoteData.svelte
@@ -1,0 +1,25 @@
+<script lang="ts">
+    import type { RemoteData } from '../lib/api';
+
+    export let data: RemoteData<unknown>;
+</script>
+
+{#if data.stage === 'pending'}
+    <slot name="pending">
+        Loading...
+    </slot>
+{:else if data.stage === 'success'}
+    <slot name="success" data={data.data}>
+        Missing success slot!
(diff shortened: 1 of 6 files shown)


gui/src/components/RemoteData.svelte
```diff
@@ -0,0 +1,36 @@
...
+  export let data: RequestLifecycle<Item>;
+</script>
+
+{#if data.stage === 'pending'}
+  <slot name="pending">Loading...</slot>
+{:else if data.stage === 'success'}
+  <slot name="success" data={data.data}>Missing success slot!</slot>
+{:else if data.stage === 'refetching'}
+  <slot name="refetching" data={data.data}>
+    <slot name="success" data={data.data}>
```
brandonbloom, line 28: if i'm reading this correctly, this is defaulting refreshing to success? intersting/neat

kendru: Yes - I just discovered this, and it seemed like a good way to handle this sort of pattern of providing implementations at the right level of specificity so that the client code can decide to handle re-fetching differently than just having data.

PR #199, opened 2021-08-16
Use CSS variables app-wide and clean up styling code

jwmza:
This swaps from in-situ coloring to `theme.css` CSS variable coloring app-wide in order to centralize the color palette and facilitate future initiatives:

- Further reduction of specific/repetitive styling, by moving towards a more minimal/efficient palette.
- Basic color theming (light/dark/black)
- Potential advanced user color themes
- Allow high leverage re-coloring at the component/parent level to theme components without replicating excessive styling code

### DIFF
--- a/gui/src/components/logs/FormattedLogMessage.svelte
+++ b/gui/src/components/logs/FormattedLogMessage.svelte
@@ -68,26 +68,21 @@
   }
 
   @keyframes blink {
-    0% {
-      visibility: hidden;
+    from {
+      opacity: 0;
+    }
+    49% {
+      opacity: 0;
     }
     50% {
-      visibility: hidden;
+      opacity: 1;
     }
-    100% {
-      visibility: visible;
+    to {
+      opacity: 1;
     }
   }
 
   :global(.exo-message-invert) {
-    background: white;
     filter: invert(1);
-  }
-
-  @media (prefers-color-scheme: dark) {
-    :global(.exo-message-invert) {
-      background: black;
-      filter: invert(1);
-    }
   }
 </style>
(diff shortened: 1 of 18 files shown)


gui/src/components/logs/FormattedLogMessage.svelte
```diff
@@ -68,26 +68,21 @@
   }
 
   @keyframes blink {
-    0% {
```
brandonbloom, line 71: Was this broken somehow? Did you try your changes in multiple browsers in Storybook? :)

brandonbloom (approved):
LGTM

gui/src/components/logs/FormattedLogMessage.svelte
jwmza: I figured it makes more sense to use `opacity` rather than `visibility` to avoid running into any unexpected behavior around the pointer events being removed for `visibility: hidden`, but since `opacity` has intermediate values between `0` and `1` it needs the `49% -> 50%` keyframes

Only tested in Chromium but will try others before merge. Opacity is very old and well supported though.

brandonbloom: So this is cartoonishly low priority, as blink is virtually unused and I added it just for fun, but...

I really think it made mode sense before. When you mess with opacity, you need to do alpha blending and stuff like that. There is some amount of time between 49% and 50%, so you're going to have some slight animation in there. It was binary, discrete, now it's scalar, continuous, even if there are sharp edges.

No need to change it back, but I don't think there was a change it in the first place.

PR #200, opened 2021-08-17
Fix logging

### LINKED ISSUE #127: [BUG] exo run handles lots of output poorly
**Describe the bug**
If a command outputs a vast amount of logs exo run doesn't appear able to stop.

**To Reproduce**
1. Create a docker compose that looks like this:
```
services:
  busy:
    image: bash
    command: "bash -c 'while true; do echo hello; done'"
```
2. Send a SIGINT or SIGTERM to the `exo run` process
3. Witness as nothing happens

**Expected behavior**
The workspace should stop, terminating the container.

My assumption is that this happens because of the very large amount of output that the script generates.

### LINKED ISSUE #162: [BUG] Initial logs are missing from exo run
**Describe the bug**
When using `exo run` the first few seconds of logs are missing.

**To Reproduce**
1. Create a docker compose file like this:
```docker-compose.yaml
services:
  test:
    image: bash
    command: "bash -c 'i=0; while true; do echo $((i++)); sleep 1; done'"
```
2. Witness this behaviour:
```
❯ dexo run docker-compose.y*
GUI available at: http://localhost:4001/#/workspaces/hk6vv1fe87qsjqn7kbb8ywaanr
Job URL: http://localhost:4001/#/jobs/t9sayfxh6031e8tjrk9f0pxq88
applying  success
  └─ updating test  success
16:08:15 test 3

16:08:16 test 4
```
Notice that the first log line printed is "3" when it should have been "0".

### LINKED ISSUE #201: [BUG] Logs in GUI are completely broken
The GUI will sometimes render one page of logs, but it does not fetch any more. Sometimes no logs are displayed.

kendru:
This addresses the issues that we have been experiencing in the UI with logs by doing the following:
- Changing the `TwoColumn` component to use named slots. The anonymous slots were causing the inner components to mount twice (and sometimes to render in the wrong place when live-updated), which in turn caused state initialization logic to run twice.
- Moving from a global store for logs to a component-based provider/consumer pattern.
- Fixing where a store being unnecessarily updated was leading to over-rendering components and requesting data too often.

Fixes #127 
Fixes #162 
Fixes #201


jwmza (approved):
LGTM - Thanks for cleaning up the formatting code and catching the slot issue as well.

gui/src/components/logs/LocalLogProvider.svelte
```diff
@@ -0,0 +1,70 @@
+<script lang="ts">
```
brandonbloom, line 1: is there a notion of a RemoteLogProvider? Seems like this could be called "LogProvider", no?

gui/src/components/logs/Logs.svelte
```diff
@@ -1,30 +1,63 @@
...
+  });
 </script>
 
-<table>
-  {#each logs as log (log.id)}
-    <tr style={log.style}>
-      <td
-        class="time"
-        on:click={() => {
-          window.alert(`Full timestamp: ${log.time.full}`);
```
brandonbloom, line 14: I missed this in @jwmza's recent PR, but it looks like 9009f192b483adc0b08a589a3ee5bf0c38ab001c undid some of the work that I did in 9009f192b483adc0b08a589a3ee5bf0c38ab001c - Why?

brandonbloom (approved):
LGTM

gui/src/components/logs/Logs.svelte
jwmza: The floating timestamp was causing some usability issues with copy/paste as well as text formatting & whitespace issues. Unfortunately an in-situ solution is not really viable due to the visual overflow of the log table component, so it needs some way to render ontop of the application in some sort of messages/alerts layer.

The other solution is a toggle which toggles all the timestamps to full timestamps, which wouldn't cause the issues listed above.

Relocated full timestamp to `window.alert` in the meantime.

PR #204, opened 2021-08-17
Minor UI improvements (buttons,...)

jwmza:
- Use styled buttons
- Refactor and extract components out of complicated pages and parent components
- Replace blue running spinner with a green running light
- Create a generic `Spinner` component


jwmza:
In addition to extracting components out of the previously much larger `ProcessList.svelte`, I've also replaced the blue running spinners with a temporary solution to #208  @brandonbloom 

image

Though we _should_ figure out some useful way to match up the colors from the logs with the colors from the process list.

Also, it's kind of strange that `[process-here]:error` uses the same color palette in logs as the normal process logs, shouldn't we make them red somehow or use an alternative styling somehow? It's kind of hard to tell they're errors.

jwmza:
This now also provides an even better `blink` than before, which only affects the text itself and uses a `steps` animation function to make a perfect 50-50 binary animation between transparent and visible text.

```
  :global(.exo-message-blink) {
    animation: blink 1s steps(1) infinite;
  }

  @keyframes blink {
    50% {
      color: transparent;
    }
  }
```

gui/src/components/Spinner.svelte
```diff
@@ -1,4 +1,8 @@
-<div />
+<script lang="ts">
+  export let inline: boolean = false;
```
brandonbloom, line 2: nit: you can use inference on the type here

```suggestion
  export let inline = false;
```

brandonbloom (approved):

PR #210, opened 2021-08-17
InitResource args now passed via newController in ComponentBase

brandonbloom:
I wanted component name deeper down in the system (should also be helpful for #192), so I made this refactor.


BenElgar (approved):
:+1:

PR #211, opened 2021-08-17
Perform shallow parsing of compose files during manifest import.

brandonbloom:


internal/manifest/compose/convert.go
```diff
@@ -5,7 +5,6 @@ import (
 	"io"
 
 	"github.com/deref/exo/internal/manifest"
-	"github.com/deref/exo/internal/providers/docker/compose"
```
brandonbloom, line 8: This PR primarily exists to break this dependency. Importing needs to preserve interpolation expressions and doesn't need to deeply understand the configuration within services or other components.

internal/manifest/compose/parse.go
```diff
@@ -1,7 +1,23 @@
...
 
 import (
-	"github.com/deref/exo/internal/providers/docker/compose"
+	"io"
+
+	"github.com/goccy/go-yaml"
 )
 
-var Parse = compose.Parse
+type Project struct {
```
brandonbloom, line 9: "Project" is apparently what Docker Compose now calls these things, tho it really means the merge of multiple compose files.

PR #214, opened 2021-08-18
Fix unsupported names in docker compose

BenElgar:


internal/manifest/manifest.go
```diff
@@ -37,6 +38,12 @@ type LoadResult struct {
 	Err      error
 }
 
+func (lr LoadResult) AddRenameWarning(originalName, newName string) LoadResult {
```
brandonbloom, line 41: This probably should be a pointer-method and not return any values.

brandonbloom (approved):

brandonbloom:
please squash commits here too - the log on `main` is becoming a mess, and it was really annoying while trying to build the changelog for the last release

PR #222, opened 2021-08-18
Do not follow symlinks in docker build

BenElgar:


brandonbloom:
aside: please squash commits

Beyond that, please say more about what was broken and why this is the correct fix. Is this related to #223?

BenElgar:
This is just a small step towards making our build do the same thing as `docker build`. Docker doesn't follow symlinks, because that would allow your build context to grow in an unbounded fashion and turn the directory tree into a potentially cyclic graph, so presumably neither should we. We already don't follow symlinked directories but this expands that to include the individual files.

brandonbloom:
Gotcha, thanks for explaining.

brandonbloom (approved):

PR #225, opened 2021-08-18
Add basic mount functionality

BenElgar:


internal/providers/docker/components/container/lifecycle.go
```diff
@@ -45,6 +46,15 @@ func (c *Container) create(ctx context.Context) error {
 		}
 	}
 
+	mounts := make([]mount.Mount, len(c.Spec.Volumes))
+	for i, v := range c.Spec.Volumes {
+		if mount, err := makeMountFromVolumeString(c.WorkspaceRoot, v); err != nil {
+			return err
+		} else {
```
brandonbloom, line 53: a few go-isms here:

1) the `else` is unnecessary and non-idiomatic.
2) this error needs context! something like `fmt.Errorf("invalid mount at index %d: %w", i, err)`

internal/providers/docker/components/container/lifecycle.go
```diff
@@ -45,6 +46,15 @@ func (c *Container) create(ctx context.Context) error {
 		}
 	}
 
+	mounts := make([]mount.Mount, len(c.Spec.Volumes))
```
brandonbloom, line 49: same nitpicky feedback as a previous PR: don't need this temporary, can directly add mounts in to the `containerCfg` object. I prefer that style for a few reasons: 1) it's one fewer mistake to make, forgetting to assign the value 2) one line shorter for every case of having to build these things up 3) eliminates any clarity about whether or not the temporary is useful anywhere else.

internal/providers/docker/components/container/mount.go
```diff
@@ -0,0 +1,61 @@
...
+	"fmt"
+	"os/user"
+	"path/filepath"
+	"strings"
+
+	"github.com/docker/docker/api/types/mount"
+)
+
+func makeMountFromVolumeString(workspaceRoot, volume string) (mount.Mount, error) {
+	// This obviously doesn't handle colons in the path well but it would appear
```
brandonbloom, line 13:
```suggestion
	// This doesn't handle colons in the path well but it would appear
```

:)

internal/providers/docker/components/container/mount.go
```diff
@@ -0,0 +1,61 @@
...
+	// that docker-compose doesn't handle those either.
+	volumeParts := strings.Split(volume, ":")
+	if len(volumeParts) > 3 {
```
brandonbloom, line 16: In order to handle colons in paths, would the correct thing to do here be to do `strings.LastIndex(volume, ":")` and split at that index?

internal/providers/docker/components/container/mount.go
```diff
@@ -0,0 +1,61 @@
...
+	source, target := volumeParts[0], volumeParts[1]
+	mountType := mount.TypeBind
+	if strings.HasPrefix(source, "./") {
+		var err error
+		source, err = filepath.Abs(filepath.Join(workspaceRoot, source))
+		if err != nil {
+			return mount.Mount{}, err
+		}
+	} else if strings.HasPrefix(source, "~/") {
+		user, err := user.Current()
```
brandonbloom, line 46: `user` should be a parameter. Consider (A) testing and (B) a future PaaS where we may run this or related code on remote machines or as `root` in a container, or something like that.

internal/providers/docker/components/container/mount_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"os/user"
+	"path"
+	"testing"
+
+	"github.com/docker/docker/api/types/mount"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMakeMount(t *testing.T) {
+	user, err := user.Current()
```
brandonbloom, line 13: heh, i hadn't gotten this far when i made the earlier comment about injecting the user as a parameter :)

brandonbloom (approved):
LGTM. All comments minor.

internal/providers/docker/components/container/lifecycle.go
BenElgar: The else is necessary so that `mount` is in scope but I changed it from 
```go
		if mount, err := makeMountFromVolumeString(c.WorkspaceRoot, v); err != nil {
			return err
		} else {
			hostCfg.Mounts[i] = mount
		}
```
to
```go
		mnt, err := makeMountFromVolumeString(c.WorkspaceRoot, v)
		if err != nil {
			return err
		}
		hostCfg.Mounts[i] = mnt
```
assuming you prefer that style.

internal/providers/docker/components/container/mount.go
BenElgar: I don't think that fixes it since the separating colon appears between two file paths. So splitting on the last colon also does the wrong thing for `./app:/du:mb/filepath`.

PR #229, opened 2021-08-18
Respect .dockerignore files

### LINKED ISSUE #224: [BUG] dockerignore not respected
**Describe the bug**
We don't currently take `.dockerignore` files into account when creating the build context.

brandonbloom:
Fixes #224


internal/providers/docker/components/container/build.go
```diff
@@ -238,3 +257,47 @@ func tarBuildContext(w io.WriteCloser, root string) (err error) {
 	}
 	return nil
 }
+
+func createIgnoreMatcher(buildContext string) (*fileutils.PatternMatcher, error) {
+	// Initialize patterns with always ignored files.
+	const ignoreBasename = ".dockerignore"
+	patterns := []string{
+		ignoreBasename,
```
BenElgar, line 265: Why are we including `.dockerignore` in the list of ignored files? Docker doesn't:

```
❯ cat Dockerfile
FROM bash
COPY . .

❯ cat .dockerignore
my-file

❯ docker run $(docker build -q .) cat .dockerignore
my-file
```

Or am I misunderstanding what's happening?

internal/providers/docker/components/container/build.go
```diff
@@ -238,3 +257,47 @@ func tarBuildContext(w io.WriteCloser, root string) (err error) {
 	}
 	return nil
 }
+
+func createIgnoreMatcher(buildContext string) (*fileutils.PatternMatcher, error) {
+	// Initialize patterns with always ignored files.
+	const ignoreBasename = ".dockerignore"
```
BenElgar, line 263: I'll note, just as an FYI, that docker also supports ignore files of the format `<dockerfile-name>.dockerignore`. Though you have to enable Buildkit for that to work. It's reasonably widely used and I've used the feature in every non-trivial docker project I've worked on, including the console.

https://stackoverflow.com/a/57774684/1142314
https://github.com/deref/deref/blob/master/registry/build.Dockerfile.dockerignore

BenElgar (approved):

internal/providers/docker/components/container/build.go
brandonbloom: I was copying what I saw in moby/moby, but I forget where. Maybe I misread it.

internal/providers/docker/components/container/build.go
brandonbloom: Do we have a ticket somewhere to support Buildkit builds?

PR #230, opened 2021-08-19
Add Dark Mode

jwmza:
# Ta-da!

image

## Gallery

image

image

image

image

image

## Todo

- [x] Inputs and code editors need to be dark.
- [ ] Fix any small issues. (e.g. button text and background both dark in some cases?)
- [x] Fix Event Listener memory leak with `onDestroy`

**Later:**
- [ ] User preferences page to manually select color theme (**auto, light, dark, black**) and synchronize across browsers/sessions with a stored preference file.
- [ ] Improve contrast for dark greys.
- [ ] Replace Event Listener JS theme switch for `TextEditor` with a singleton store to access theme within TS
- [ ] **Black** and **Dark** themes to give users an option between pure-black `#000000` OLED-friendly dark theme and the more common dark-grey backgrounds dark theme.

### DIFF
--- a/gui/src/lib/color.ts
+++ b/gui/src/lib/color.ts
@@ -13,13 +13,21 @@ export const textHoverColor = (deg: number) => `hsl(${deg}, 95%, 20%)`;
 export const bgColor = (deg: number) => `hsl(${deg}, 65%, 92%)`;
 export const bgHoverColor = (deg: number) => `hsl(${deg}, 65%, 87%)`;
 
+export const darkTextColor = (deg: number) => `hsl(${deg}, 75%, 75%)`;
+export const darkTextHoverColor = (deg: number) => `hsl(${deg}, 75%, 90%)`;
+export const darkBgColor = (deg: number) => `hsl(${deg}, 80%, 6%)`;
+export const darkBgHoverColor = (deg: number) => `hsl(${deg}, 80%, 12%)`;
+
 // This computes a HTML style attribute string for colored logs.
 export const logStyleFromHash = (s: string) => {
   const d = hashDegree(s);
   // Combine styles into one string.
-  return `--log-color:${textColor(d)};--log-bg-color:${bgColor(
-    d,
-  )};--log-hover-color:${textHoverColor(d)};--log-bg-hover-color:${bgHoverColor(
-    d,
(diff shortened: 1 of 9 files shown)


jwmza:
Please let me know of any missing variables / visual bugs. I've made a brief list above of what I'm currently aware of.

brandonbloom (approved):

gui/src/lib/color.ts
```diff
@@ -13,13 +13,21 @@ export const textHoverColor = (deg: number) => `hsl(${deg}, 95%, 20%)`;
 export const bgColor = (deg: number) => `hsl(${deg}, 65%, 92%)`;
 export const bgHoverColor = (deg: number) => `hsl(${deg}, 65%, 87%)`;
 
+export const darkTextColor = (deg: number) => `hsl(${deg}, 75%, 75%)`;
+export const darkTextHoverColor = (deg: number) => `hsl(${deg}, 75%, 90%)`;
+export const darkBgColor = (deg: number) => `hsl(${deg}, 80%, 6%)`;
+export const darkBgHoverColor = (deg: number) => `hsl(${deg}, 80%, 12%)`;
+
 // This computes a HTML style attribute string for colored logs.
 export const logStyleFromHash = (s: string) => {
```
brandonbloom, line 22: Ideally, this would take some sort of `theme: Theme` argument, so that all the color handling logic for a theme can be centralized, rather than spread out all over the place. Not pressing now though.

jwmza: This kinda causes some issues with spreading out how themes are handled @brandonbloom 

Although, I think it may be necessary anyways. Looking at the Monaco editor, it looks like the theme needs to be set by the JS code and cannot be set simply by a CSS media query :/

jwmza: Never mind I got the Monaco editor working

brandonbloom (approved):
LGTM now, assuming you feel that this is safe to merge before the other improvements are made.

jwmza:
Monaco theme finally fixed

image
image

jwmza:
Okay I'm merging this but no guarantees of perfect coverage or lack of visual bugs :) will be fixing things from now until the next major release

PR #236, opened 2021-08-19
Fix logs newline bug

jwmza:


BenElgar (approved):

gui/src/components/logs/Logs.svelte
```diff
@@ -27,6 +27,12 @@
       logViewport.scrollTop = logViewport.scrollHeight;
     }
   });
+
+  // This trims the final newline character at the end of each log message,
```
brandonbloom, line 31: What final newline? Seems like this is fixing a symptom, not the bug: The log message should not contain a newline in the first place.

BenElgar: Why wouldn't it include a newline character? If a program emits `hello\nbye` we should presumably record that and then, when displaying the logs, decide how best to render that output?

brandonbloom: The logs are not stored or transmitted as streams of text, but rather as events. New lines are parsed, so hello\nbye would become two events. The intention is that each event’s message ends with an implied newline.

BenElgar: Sure but I would still expect the concatenation of the contents of each event to be equal to the original output. The new lines are just our mechanism of breaking up the stream of text. Otherwise how does one tell if, for example, the line ended with `\r\n` or just `\n`?

brandonbloom: Is there a reason to recreate the original output precisely? I don’t think there is. Syslog has a few different encodings, the most popular of which uses line terminators to delimit messages and mandates Unix line endings. Windows line endings are lost. I think that’s to be expected.

BenElgar: To me the question is why do anything else? We're already preserving the original output — why change that?

I also though it was how other tools operated because I remember having to do something similar when I was working with DataDog logs (I think).

jwmza: We found this to be an issue with final newlines in the raw HTML DOM text on @BenElgar 's machine, combined with the whitespace preserving `white-space` property we use.

My thinking is it could either be (a) an issue with browsers running in Linux machines, because my FireFox (which runs in Windows) didn't have the issue, or (b) some other issue due to Ben's specific machine setup. Maybe it's Svelte's fault, maybe it's the fault of the browser, or the OS, or something else. I'm not sure how we can effectively debug this.

Doing this was the easiest and cleanest way to fix the issue (without creating additional issues) we could find.

PR #238, opened 2021-08-19
Compose host resolution

### LINKED ISSUE #220: [BUG] docker-compose networking doesn't work
**Describe the bug**
docker compose makes services accessible to each other via a hostname equal to the name of the service. This doesn't work with `exo run`.

```docker-compose.yaml
services:
  echo:
    image: hashicorp/http-echo
    command: '-text="hello world"'
  t1:
    image: curlimages/curl
    command: sh -c 'while true; do curl http://echo:5678; sleep 1; done'
```
`docker-compose up` correctly prints `hello world` whilst `exo run` fails to resolve the host:
 ```
curl: (6) Could not resolve host: echo
```

### PR DESCRIPTION
**Pull request by kendru**
This PR updates the docker-compose networking such that the containers defined by each docker-compose service uses any configured networks as well as a default network that is implicitly created for each compose project.

Fixes #220

Applying a new project does not always start the containers because they cannot attach to the network if they are created before the network. Similarly, the workspace cannot be destroyed in a single command because the containers have to be removed before the network can be removed. Addressing these dependency issues will be done in #237.

### CODE REVIEW

File: .dev/config.toml [hunk 1]
```diff
@@ -1,5 +1,8 @@
 httpPort = 4001
 
+[client]
+url = "http://localhost:4001"
```
kendru, line 4: This is not related, but it will prevent `dexo` from trying to start a daemon unless running `dexo daemon` explicitly.

internal/providers/docker/components/container/lifecycle.go
```diff
@@ -229,8 +232,20 @@ func (c *Container) create(ctx context.Context) error {
...
-		//EndpointsConfig map[string]*EndpointSettings // Endpoint configs for each connecting network
+		EndpointsConfig: make(map[string]*network.EndpointSettings), // Endpoint configs for each connecting network
+	}
+	for _, networkName := range c.Spec.Networks {
+		networkCfg.EndpointsConfig[networkName] = &network.EndpointSettings{
+			Aliases: []string{c.ComponentID, c.ComponentName},
+		}
 	}
+
+	if jsonBytes, err := json.MarshalIndent(networkCfg, "", "  "); err == nil {
```
brandonbloom, line 243: leftovers?

internal/manifest/compose/convert.go
```diff
@@ -8,50 +8,144 @@ import (
...
 		}
+		n := network.ToMap()
+
+		// If a `name` key is specified in the network configuration (usually used in conjunction with `external: true`),
+		// then we should honor that as the docker network name. Otherwise, we should set the name as
+		// `<project_name>_<network_key>`.
+		if _, ok := n["name"]; !ok {
+			n["name"] = i.prefixedName(name)
+		}
+		networksByComposeName[originalName] = n["name"].(string)
```
brandonbloom, line 72: This could potentially panic.

internal/manifest/compose/convert.go
```diff
@@ -8,50 +8,144 @@ import (
...
-		if name != originalName {
+		if originalName != name {
 			res = res.AddRenameWarning(originalName, name)
 		}
+
+		v := volume.ToMap()
+		if _, ok := v["name"]; !ok {
+			v["name"] = i.prefixedName(name)
+		}
+		volumesByComposeName[originalName] = v["name"].(string)
```
brandonbloom, line 45: This can potentially panic. Would have been nice to modify the types to get some validation during unmarshalling.

Issue #217, opened 2021-08-18
[BUG] Docker commands parsed badly


BenElgar:
**Describe the bug**
If you try and run an docker-compose file like this:

```yaml
services:
  t1:
    image: bash
    command: "echo hello && echo bye"
```

A component state like this will be written to the exo `state.json`:
```
        "6vcn55x4egy88ppmhr2deqbm3c": {
          "name": "t1",
          "type": "container",
          "spec": "image: bash\ncommand: echo hello \u0026\u0026 echo bye\n",
          "state": "",
          "created": "2021-08-18T11:00:04.4356297Z",
          "initialized": null,
          "disposed": null
        },
```

This then will fail to unmarshal:
```
❯ dexo run docker-compose.y*
2021/08/18 12:00:04 Telemetry disabled; not collecting statistics.
GUI available at: http://localhost:4001/#/workspaces/da9c7tmnvmmpvq9tpkhw69sdg8
Job URL: http://localhost:4001/#/jobs/gjh750xzg9dsrspw8dgs8kv60r
applying unmarshalling spec: 1:12: && is not a valid word
✓ ├─ adding test
⨯ └─ adding t1 unmarshalling spec: 1:12: && is not a valid word
Error: applying manifest: job failure: unmarshalling spec: 1:12: && is not a valid word
applying manifest: job failure: unmarshalling spec: 1:12: && is not a valid word
```

This is particularly bad because the state file is now corrupted and the error will continue to happen even if the command is changed to no longer include the offending characters.

```
❯ dexo rm t1
2021/08/18 12:06:51 Telemetry disabled; not collecting statistics.
Job URL: http://localhost:4001/#/jobs/v1kny9fkgksswmhkm238dskjt8
deleting unmarshalling spec: 1:12: && is not a valid word
⨯ └─ t1 unmarshalling spec: 1:12: && is not a valid word
Error: job failure: unmarshalling spec: 1:12: && is not a valid word
job failure: unmarshalling spec: 1:12: && is not a valid word
```

brandonbloom:
The fix in #239 is incomplete. The compose reference says the correct behavior is defined as per the CMD directive, defined in the Dockerfile spec. Here:
https://docs.docker.com/engine/reference/builder/#cmd

Looks like the right thing to do is to accept either a individual string or an array of strings. If an individual string, invoke the [SHELL](https://docs.docker.com/engine/reference/builder/#shell) directly.

PR #239, opened 2021-08-20
Accept more formats when parsing docker commands

BenElgar:
This is still wrong but a bit less wrong.

### DIFF
--- a/internal/providers/docker/compose/command.go
+++ b/internal/providers/docker/compose/command.go
@@ -4,6 +4,11 @@ import (
 	"github.com/mattn/go-shellwords"
 )
 
+// FIXME: this incorrectly treats the "command" passed to docker as a
+// meaningful shell command that can be parsed into "shell words". In reality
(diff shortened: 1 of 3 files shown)


internal/providers/docker/compose/command.go
```diff
@@ -1,7 +1,14 @@
...
-import "mvdan.cc/sh/v3/shell"
+import (
+	"github.com/mattn/go-shellwords"
+)
 
+// FIXME: this incorrectly treats the "command" passed to docker as a
+// meaningful shell command that can be parsed into "shell words". In reality
+// this is an arbitrary string that is passed to the entrypoint of the
+// container as an argument. It could be a shell command but it could also be
+// a python program or a Dan Brown novel.
```
brandonbloom, line 11: Rather than a joke here, would have been nice to explain the problem with reference to the spec and or what to do about it. I've re-opened #217 and will add context there.

PR #241, opened 2021-08-20
New config pages

jwmza:
This is going to add a new set of **CRU**~D~ pages and components for configuration, specifically aimed at:

- Component type selection
- Process creation and modification
- Docker container creation and modification
- Timer creation and modification
- User preferences config


jwmza:
This now adds a functional theme switch capability backed by Svelte's Stores and synced+loaded from the `localStorage` API.

I believe I've fixed any issues with switching, including the log colors and Monaco editor theme. If anybody finds new issues please ping me.

The primary remaining questions regarding theme is:

- Should we use an OS storage config file like the existing config (e.g. for telemetry) for user GUI preferences, including theme/ligatures/fonts/etc, or should we use `localStorage` as implemented here? I don't see an _obvious_ answer, but the key difference is that if we set these preferences in the backend they will persist in different browsers and can be accessed within the backend as well. I'm thinking we put all preferences in the same config file.
- The `theme.css` file is getting pretty messy. I can think of a few things to clean it up slightly, but it doesn't even have dark/black different modes yet and overall I'm thinking it should probably be generated from a cleaner file that defines each theme variable once instead of forcing manual synchronization of 4-5 theme variable declaration chunks. I'm thinking of a single JSON file with, for example:
```
...,
"link-color": {
  light: "#00f",
  dark: "#0f0",
  black: "#f00"
},
...
```
resulting in generated code chunks:
```
:root, body {
  ...
  --link-color: #00f;
  ...
}

@media (prefers-color-scheme: dark) {
  :root, body {
    ...
    --link-color: #f00;
    ...
  }
}

body.light {
  ...
  --link-color: #00f;
  ...
}

body.black {
  ...
  --link-color: #f00;
  ...
}

body.dark {
  ...
  --link-color: #0f0;
  ...
}
```

jwmza:
Theme file generation working

jwmza:
@kendru this still needs some cleanup in the preferences config page (i.e. removing the static stuff at the bottom) but I think it's a good idea to try to merge this with mainly the theme stuff + preferences page first instead of making it a monster PR that includes the docker container config / create page and dynamic config input components and such

script/gen-theme.js
```diff
@@ -0,0 +1,172 @@
...
+  "grey-b-color": ["#bbbbbb", "#444444", "#444444"],
+  "grey-c-color": ["#cccccc", "#333333", "#333333"],
+  "grey-d-color": ["#dddddd", "#222222", "#222222"],
+  "grey-e-color": ["#eeeeee", "#111111", "#111111"],
+  "grey-e7-color": ["#e7e7e7", "#0c0c0c", "#0c0c0c"],
+  "grey-f9-color": ["#f9f9f9", "#070707", "#070707"],
+  "layout-bg-color": ["#cccccc", "#222222", "#222222"],
+  "secondary-bg-color": ["#f5f5f5", "#0a0a0a", "#0a0a0a"],
+  "primary-bg-color": ["#ffffff", "#000000", "#000000"],
+  /* Contextual color */
```
brandonbloom, line 27: nit: use `//` comments in js/ts files unless commenting intra-line.

script/gen-theme.js
```diff
@@ -0,0 +1,172 @@
+var fs = require("fs");
+
+// name: [light, dark, black]
+
+var themeVariables = {
```
brandonbloom, line 5: `let` and `const` are available in js files & have been in node for a long time.

script/gen-theme.js
```diff
@@ -0,0 +1,172 @@
...
+    "var(--dark-log-hover-color)",
+  ],
+  "log-bg-hover-color": [
+    "var(--light-log-bg-hover-color)",
+    "var(--dark-log-bg-hover-color)",
+    "var(--dark-log-bg-hover-color)",
+  ],
+};
+
+function themeDefinition(theme) {
```
brandonbloom, line 144: nit: prefer arrow functions unless you need mutual recursion or other obscure reasons for using `function`

script/gen-theme.js
```diff
@@ -0,0 +1,172 @@
...
+  return Object.entries(themeVariables)
+    .map((entry) => {
+      return `--${entry[0]}: ${entry[1][theme]};`;
```
brandonbloom, line 147: The generated code would be much easier to read if there were a `\n` on the end of each of these lines.

brandonbloom (commented):
There is too much going on at once in this PR to review it properly. I'd go commit-by-commit, but there are 25 of them.

Can we separate out the theme generation stuff independently and review that first as its own PR?

Is the docker container UI stuff behind a feature flag or is it working well enough to be used? Or would merging this introduce some incomplete/broken features?

jwmza:
I'll fix those nits @brandonbloom 

Re: the generated file, I'm not sure what the right ideal to strike between size and legibility is but I guess there's little harm done to size in making it more legible with newlines. Will fix.

Re: new component UI code, it's commented out in the Preferences page and is very WIP but introduces no breaking changes.
The final thing will require a mix of the type of code in the new process page right now + cleaner, componentized versions of that new code. Everything incomplete is commented out.

Notably, the new component page is "complete" but only contains one option, "New process", at the moment, so I figure we should probably effectively skip over the page in routing for now.

jwmza:
The theme generation and manual theme pref switching stuff is interlinked and I don't think it would make much sense to separate them out. In theory it's possible but having either without the other is unnecessarily complicated.

The only stuff in here that should've been separated out in hindsight from the start is the new component page and the two smaller PRs, plus some visual fixes.

jwmza:
I believe the only important files with meaningful changes other than the theme generation stuff and new graphics are the following:

- `Layout.svelte`
- `WorkspaceNewComponent.svelte`
- `NewProcess.svelte`
- `Preferences.svelte`

gui/src/lib/gui-theme.ts
```diff
@@ -0,0 +1,34 @@
...
+    dark: () => localSyncedSet('dark'),
+    black: () => localSyncedSet('black'),
+  };
+}
+
+export const guiTheme = createGuiTheme();
+
+export const setGuiTheme = (theme: string) => {
+  for (const option of ['auto', 'light', 'dark', 'black']) {
+    if (theme === option) {
```
brandonbloom, line 28: this add/remove pattern can be simplified to use `toggle`. See https://developer.mozilla.org/en-US/docs/Web/API/Element/classList

```typescript
document.body.classList.toggle(option, theme === option)
```

gui/src/lib/gui-theme.ts
```diff
@@ -0,0 +1,34 @@
...
+    auto: () => localSyncedSet('auto'),
+    light: () => localSyncedSet('light'),
+    dark: () => localSyncedSet('dark'),
+    black: () => localSyncedSet('black'),
+  };
+}
+
+export const guiTheme = createGuiTheme();
+
+export const setGuiTheme = (theme: string) => {
```
brandonbloom, line 26: I was really confused by this name. It's not really "setGuiTheme" so much as it is "renderGuiTheme" or "applyGuiTheme". Aside, "gui theme" seems redundant - why not just "theme"?

Everywhere you call `auto()` or `light()` or similar, you then immediately call `setGuiTheme`. Could that be automatic as part of `localSyncedSet`?

Actually, moreover, this probably doesn't need to be here at all. The guiTheme export is already subscribable, so Layout should add a reactive block that re-runs each time the theme changes.

gui/src/lib/gui-theme.ts
```diff
@@ -0,0 +1,34 @@
...
+  const { subscribe, set } = writable(ls);
+
+  const localSyncedSet = (s: string) => {
+    set(s);
+    localStorage.setItem(guiThemeLocalStorageKey, s);
+  };
+
+  return {
+    subscribe,
+    auto: () => localSyncedSet('auto'),
```
brandonbloom, line 17: why are there 4 methods here rather than `set(value: ThemeName)` or something like that?

script/gen-theme.js
```diff
@@ -0,0 +1,193 @@
+const fs = require("fs");
+
+// Run `/path/to/exo $ node sript/gen-theme.js` to generate output.
```
brandonbloom, line 3: typo: sript

script/gen-theme.js
```diff
@@ -0,0 +1,193 @@
...
+// Format a block of theme variables for root and child selectors.
+const themeBlock = (selector, theme) => `body.${selector} {
+${themeDefinition(theme)}
+}
+body.${selector} * {
+${childDefinition(theme)}
+}`;
+
+// Format the entire theme file using the above helpers.
+const out = `/* Generated file. DO NOT EDIT. */
```
brandonbloom, line 180: not that it matters much for a small code generator like this, but in the future, you should probably perform code generation imperatively by doing `out += ...` instead of using string interpolation to build up a large string. Otherwise, you wind up copying the strings over and over and over again as you iteratively rebuild immutable strings up the tree of interpolations.

gui/src/components/Layout.svelte
```diff
@@ -1,27 +1,45 @@
...
+  import FeedbackSVG from './mono/FeedbackSVG.svelte';
+  import PreferencesSVG from './mono/PreferencesSVG.svelte';
+  import { guiTheme, setGuiTheme } from '../lib/gui-theme';
+  import { onMount } from 'svelte';
+  import * as router from 'svelte-spa-router';
 
-  const goHome = () => {
-    router.push('#/');
-  };
+  onMount(() => {
```
brandonbloom, line 10: I discuss `setGuiTheme` elsewhere, so go see that comment.

This is where you'd want something like:

```
$: {
  doWhateverTo(document.body.classList, $guiTheme);
}
```

gui/src/components/Layout.svelte
```diff
@@ -1,27 +1,45 @@
...
 
 <main>
   <nav>
     <header>
-      <NavbarButton title="Home" on:click={goHome}>
-        <img src="/deref-rounded-icon.png" alt="Deref" height="24px" />
+      <NavbarButton
+        title="Home"
+        on:click={() => {
+          router.push('#/');
```
brandonbloom, line 21: Maybe `NavbarButton` should have a `href` property, which would clean up all these `on:click` handlers.

gui/src/components/TextEditor.svelte
```diff
@@ -36,8 +37,11 @@
     const mqList = window.matchMedia('(prefers-color-scheme: dark)');
 
     const handleThemeChange = (e: MediaQueryListEvent) => {
-      setTheme(e.matches);
-      console.log('test color listener');
+      setTheme(
+        (e.matches && $guiTheme === 'auto') ||
```
brandonbloom, line 41: seems like `guiTheme` is actually doing double duty: one for the selection and one for the effective theme (ie having resolved auto to light, or dark, or black) - can actually hoist that up in to a reactive value at the store level. I might call them something like `selectedTheme` and `effectiveTheme` and then shorten the more commonly/widely used one (effectiveTheme) to just `theme`.

gui/src/pages/Preferences.svelte
```diff
@@ -0,0 +1,113 @@
...
+            }}
+            inset={$guiTheme === 'black'}
+          >
+            Black
+          </Button>
+        </div>
+      </div>
+      <hr />
+      <p>
+        More preferences will be available here soon. In the meantime, please
```
brandonbloom, line 70: Not a fan of the prose here. Feels very "UNDER CONSTRUCTION". Seems like we can simply drop this entire `<p>` for now.

brandonbloom (approved):
OK, this is now understandable in isolation. Thank you for breaking it down.

The PR overall LGTM, but 37 work-in-progress commits is a fair bit messy. Please use "squash and merge":

image

gui/src/lib/gui-theme.ts
```diff
@@ -0,0 +1,34 @@
+import { writable } from 'svelte/store';
+
+export const guiThemeLocalStorageKey = 'io.deref.exo/gui-theme';
+
+function createGuiTheme() {
+  const ls = localStorage.getItem(guiThemeLocalStorageKey) || 'auto';
```
brandonbloom, line 6: generally, i don't think we should use local storage. instead, we should strive to keep all of the config in the `config.toml` file.

@kendru how hard would it be for us to add `git config ...` -style mutation operations? Expose those to the UI?

Issue #237, opened 2021-08-19
[BUG] Containers do not start when applying a docker-compose project


kendru:
**Describe the bug**
When a `docker-compose.yml` file is imported, exo creates containers for each of the services as well as (at least one) network to attach the containers to.  If the containers end up getting created before the network is available, they will not start automatically. Similarly, when destroying a project, we try to destroy all components in parallel, and destroying the network fails because there are still containers attached to it.

Components should have some sort of dependency ordering that should be respected when starting/stopping/removing everything in a workspace.

brandonbloom:
See also #171

Issue #171, opened 2021-08-13
[FEATURE] Component dependencies


brandonbloom:
compose files supports a `depends_on` option, documented here: https://docs.docker.com/compose/compose-file/compose-file-v3/#depends_on and here: https://github.com/compose-spec/compose-spec/blob/master/spec.md#depends_on

As of https://github.com/deref/exo/commit/8c40b5612d7338dc59db4b7caf39dc03f27c885e we load compose files and ignore this configuration. Ignoring this configuration is reasonable, since well-behaved services need to fail gracefully if their dependencies are temporarily unavailable. There is prior art for this as well, since depends_on is not supported in Docker's Swarm mode.

However, we may want dependencies for other reasons. For instance, interactive `start` operations should also start their dependencies by default.

If we do support `depends_on`, we should do so in a more general way, as this affects orchestration in exo beyond just docker.

- We should implement proper yaml unmarshalling for both the short and long syntax of compose. See the other examples, such as `PortMapping` and `Dictionary` in that package.
- During the `compose.Convert` process, we should produce dependencies in whatever way we model that in the exo manifest. Lots of design work needed here, but I have some ideas worth discussing.

PR #242, opened 2021-08-21
Dependency graph

kendru:
This work is a prerequisite to #171 and #237. It adds a dependency graph that can represent an arbitrary DAG (cycles are prevented when the graph is being built) and can return its nodes sorted topologically. Additionally, it keeps track of groups of nodes at the same "level" of the graph (i.e. nodes that have no interdependencies) and can return these groups in topological sort order as well. I plan to use this to build an execution plan for starting and stopping components such that everything that can start in parallel does start in parallel.

### DIFF
--- a/internal/deps/graph.go
+++ b/internal/deps/graph.go
@@ -1,23 +1,57 @@
 package deps
 
 import (
+	"bytes"
 	"errors"
+	"fmt"
 )
 
+type nodeset map[interface{}]struct{}
+
 // Semantically, this is Map<Node, Set<Node>>.
-type depmap map[interface{}]map[interface{}]struct{}
+type depmap map[interface{}]nodeset
 
 type Graph struct {
 	// Maintain dependency relationships in both directions.
-	// `dependencies` tracks child -> parent, and `dependents` tracks parents -> children.
+	// `dependencies` tracks child -> parents, and `dependents` tracks parent -> children.
 	dependencies, dependents depmap
+	nodes                    nodeset
 }
 
 func New() *Graph {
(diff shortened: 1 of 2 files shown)


brandonbloom:
I didn't look too carefully at the code, since it's totally isolated from the rest of the code & therefore low risk. The general need for layered topo sort makes sense to me. Though I'm curious more about the designs for #171 and #237.

My assumption is that the `apply` implementation will manually build a dependency graph of tasks to perform, toposort that, then execute the layers in parallel. I don't quite yet know how we'd build the dependency graph in an extensible way, but we can probably just violate encapsulation of the component specs for now, since we have a pretty limited set of supported types. Curious what you've got in mind.

kendru:
@brandonbloom My plan is to build the graph in `Workspace.Apply()` then translate that to an "execution plan" of tasks (which will require parallel and sequential task combinators that I'll create next). The concept of dependencies between components seems pretty fundamental to a component, so I was planning to lift it to the `ComponentDescription`. Since we will not have IDs for the components until they are created, I am debating whether to pull ID generation up into `Apply()` or to require that components be uniquely identified by name and type and referencing dependencies that way.

Also, I know that we will eventually have the need to express abstract dependencies like "component 123 depends on some `postgres` database", but that is not the problem that I am solving here. This is solely intended to be used to resolve concrete dependencies, and it could be used as a building block for resolving abstract dependencies.

brandonbloom:
Gotcha. I say we extract the ID allocation out of the core of the create codepath & move it into "apply". Seems the simplest approach for now.

PR #243, opened 2021-08-23
Fix duplicate newlines

BenElgar:


jwmza:
@BenElgar what is this?

Is this the same issue that you had in the GUI last week fixed as well in the CLI?

brandonbloom (changes requested):
This is the wrong fix. I'm pretty confident that https://github.com/deref/exo/pull/236#discussion_r692529466 is the wrong fix as well.

Instead of merging this, let's undo #236 and go fix this issue at the source, which is that the trailing newline should not be making it in to the logs database in the first place.

brandonbloom:
@jwmza, @BenElgar is there still an issue here? Would like to avoid having PRs sitting around for too long.

jwmza:
Will need to defer to @BenElgar here because to my understanding he was the only one having this issue on his system

BenElgar:
Closing in favour of https://github.com/deref/exo/pull/323/files

PR #244, opened 2021-08-23
Support for workspace .env file

BenElgar:
This adds support for a workspace having a `.env` file. At the moment there is no support for per-component environments.

### DIFF
--- a/internal/core/server/workspace.go
+++ b/internal/core/server/workspace.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"os"
 	"path"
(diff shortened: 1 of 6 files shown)


internal/core/server/workspace.go
```diff
@@ -263,15 +270,32 @@ func (ws *Workspace) newController(ctx context.Context, desc api.ComponentDescri
 }
 
 // TODO: Use workspace-defined environments, rather than ambient unix environment.
```
brandonbloom, line 272: Is this comment outdated now?

internal/core/server/workspace.go
```diff
@@ -263,15 +270,32 @@ func (ws *Workspace) newController(ctx context.Context, desc api.ComponentDescri
...
-		parts := strings.SplitN(assign, "=", 2)
-		key := parts[0]
-		val := parts[1]
-		env[key] = val
-	}
-	return env
+func (ws *Workspace) getEnvironment(ctx context.Context) (map[string]string, error) {
+	envPath, err := ws.resolveWorkspacePath(ctx, ".env")
+	if err != nil {
+		return nil, fmt.Errorf("could not resolve path to .env file: %w", err)
```
brandonbloom, line 276: the convention that the stdlib and core contributor examples tends to use is something more like `resolving path to .env file: %w`

Issue #219, opened 2021-08-18
[FEATURE] Support extended volumes syntax


BenElgar:
A docker compose file like the following doesn't work because we don't support the extended volumes syntax:
```
services:
  web:
    image: bash
    volumes:
      - type: volume
        source: mydata
        target: /data
        volume:
          nocopy: true
      - type: bind
        source: ./static
        target: /opt/app/static
volumes:
  mydata:
```

You get an error of the following form:
```
applying manifest: job failure: unmarshalling spec: cannot unmarshal map[string]interface {} into Go struct field Spec.Volumes of type string
```

kendru:
I am picking this one up because it is directly in the path of what I am doing for #215

PR #247, opened 2021-08-24
Adopt compose containers

### LINKED ISSUE #215: [BUG] Conflicts with docker-compose when using same container name
**Describe the bug**
When using the `container_name` field in a docker compose file exo will conflict with docker compose. If docker compose was run first you get an error like the following:

```
❯ dexo run docker-compose.y*
2021/08/18 11:16:10 Telemetry disabled; not collecting statistics.
GUI available at: http://localhost:4001/#/workspaces/f8t3hya2weawagt5egnp79kgag
Job URL: http://localhost:4001/#/jobs/3hn11crghc87sgrmanyavd93r4
applying adding replacement "kibana": creating container: Error response from daemon: Conflict. The container name "/kib" is already in use b
✓ ├─ updating elastic
⨯ ├─ updating logstash adding replacement "logstash": creating container: Error response from daemon: Conflict. The container name "/log" is
⨯ ├─ updating kibana adding replacement "kibana": creating container: Error response from daemon: Conflict. The container name "/kib" is alre
⨯ └─ updating elasticsearch adding replacement "elasticsearch": creating container: Error response from daemon: Conflict. The container name
Error: applying manifest: job failure: adding replacement "kibana": creating container: Error response from daemon: Conflict. The container name "/kib" is already in use by container "084b003c1ade64673d1a87e487f52a33fec4c9ed9efdf8714e4edfe84a983364". You have to remove (or rename) that container to be able to reuse that name.
applying manifest: job failure: adding replacement "kibana": creating container: Error response from daemon: Conflict. The container name "/kib" is already in use by container "084b003c1ade64673d1a87e487f52a33fec4c9ed9efdf8714e4edfe84a983364". You have to remove (or rename) that container to be able to reuse that name.
```

If `exo` was run first you get the following error:
```
❯ docker-compose up
Creating es ... error

ERROR: for es  Cannot create container for service elasticsearch: Conflict. The container name "/es" is already in use by container "b10358e50f984391c144f4193267d1dd89407e6c08829c4103b528d1f3ff1e8f". You have to remove (or rename) that container to be able to reuse that name.

ERROR: for elasticsearch  Cannot create container for service elasticsearch: Conflict. The container name "/es" is already in use by container "b10358e50f984391c144f4193267d1dd89407e6c08829c4103b528d1f3ff1e8f". You have to remove (or rename) that container to be able to reuse that name.
ERROR: Encountered errors while bringing up the project.
```

This is fixed with `docker container prune` or by removing the named containers.

I can see an argument that this isn't a bug but the correct behaviour — but it conflicts with our goal of wanting to be a drop in replacement.

kendru:
This PR fixes #215 as well as #219.

The main change is that the import process for a Docker Compose manifest no longer uses the untyped `Project` struct, but it uses the `Compose` struct from `github.com/deref/exo/internal/providers/docker/compose` instead. The main result of this is that the spec for a container now has extra fields other than the slightly-mangled fields from the source compose file that it had previously. The main motivation for doing this was to avoid duplicating parsing and validation logic. Additionally, we now check to see if there are existing volumes, networks, or containers that were previously managed by Compose. If there were containers, we remove them, and if there were networks or volumes, we "adopt" them as exo components.

This process raises some questions about the best approach for adopting existing resources. See the `ADOPT COMPOSE RESOURCES` note for more details.

Unfortunately I did not realize that I had a couple of changes from #237 on this branch, which add a `DependsOn` field to the `Component` struct. Hopefully it does not add much extra noise, but if it does, I can remove those changes from this branch and work with them separately.


internal/manifest/compose/parse.go
```diff
@@ -6,6 +6,7 @@ import (
 	"github.com/goccy/go-yaml"
 )
 
+// TODO: Consider whether this representation is useful or if we can just use `compose.Compose`,
```
brandonbloom, line 9: it's trivial to recreate - so if you're not using it now, just go ahead and delete it

internal/manifest/compose/convert.go
```diff
@@ -97,55 +103,77 @@ func (i *Loader) convert(project *Project) manifest.LoadResult {
...
+		res.Manifest.Components = append(res.Manifest.Components, component)
 	}
 
 	return res
 }
 
-func (i *Loader) prefixedName(name string) string {
-	return fmt.Sprintf("%s_%s", i.ProjectName, name)
+func (i *Loader) prefixedName(name string, suffix string) string {
+	var out bytes.Buffer
```
brandonbloom, line 169: FYI: https://medium.com/@thuc/8-notes-about-strings-builder-in-golang-65260daae6e9

internal/manifest/manifest.go
```diff
@@ -21,9 +21,15 @@ type Manifest struct {
...
-	Spec string `hcl:"spec"` // TODO: Custom unmarshalling to allow convenient json representation.
+	Name      string         `hcl:"name,label"`
+	Type      string         `hcl:"type,label"`
+	Spec      string         `hcl:"spec"` // TODO: Custom unmarshalling to allow convenient json representation.
+	DependsOn []ComponentRef `hcl:"depends_on"`
+}
+
+type ComponentRef struct {
+	Name string `hcl:"name"`
+	Type string `hcl:"type"`
```
brandonbloom, line 32: As far as I can tell, you never read out this `Type` field, but maybe I missed it somewhere? Since there is only one namespace for all types, I don't think ComponentRef should have a Type field at all. Is there a reason you added it that I'm missing?

internal/providers/docker/components/volume/lifecycle.go
```diff
@@ -2,22 +2,34 @@ package volume
...
+		// TODO: Determine whether the existing volume is compatible with the spec.
+		return &core.InitializeOutput{}, nil
+	}
+
 	labels := v.Spec.Labels.WithoutNils()
 	for k, v := range v.GetExoLabels() {
 		labels[k] = v
 	}
 
+	fmt.Printf("Creating Volume %q\n", v.Spec.Name)
```
brandonbloom, line 27: leftovers

internal/providers/docker/compose/yaml.go
```diff
@@ -160,12 +162,114 @@ type Service struct {
...
 	User string `yaml:"user"`
 	// TODO: userns_mode
-	Volumes []string `yaml:"volumes"` // TODO: support long syntax.
+	Volumes []VolumeMount `yaml:"volumes"`
 	// TODO: volumes_from
 
 	WorkingDir string `yaml:"working_dir"`
 }
 
+type VolumeMount struct {
```
brandonbloom, line 171: for sub structures with substantial parsing logic, i've been breaking those out in to their own files. this file probably should be called `project.go` instead of `yaml.go`, but I just never got around to renaming it as the contents outgrew the file :)

brandonbloom (approved):
Previously discussed caveats aside, this LGTM.

PR #266, opened 2021-08-29
docker compose interpolation

### LINKED ISSUE #265: [BUG] Docker compose interpolation not supported
See https://github.com/compose-spec/compose-spec/blob/master/spec.md#interpolation

See also: #181

brandonbloom:
Closes #265


brandonbloom:
@kendru This is still totally unfinished/untested, but did enough of it to get the intended shape of the solution in place. There are two key ideas:

1) two pass decoding. First pass decodes to generic data structures. Interpolation happens to that data structure, then it's rencoded to yaml and a second pass decodes in to the strongly typed structs. (Ideally the rencoding step would be unnecessary, but the yaml library we're using doesn't expose the ability to unmarshal directly from interface values)
2) flesh out the import-level shallow parsing types a bit more. Building on #211, which was implicitly reverted as part of the dependencies effort. This relies on the `MapSlice` and `inline` tag directive, but I don't actually know if that works at all.

internal/providers/docker/compose/interpolate/interpolate.go
```diff
@@ -0,0 +1,45 @@
...
+	err  error
+}
+
+func (r *interpolator) setErr(err error) {
+	if r.err != nil {
+		r.err = fmt.Errorf("at %v: %w", r.path, err)
+	}
+}
+
+func (r *interpolator) interpolate(v interface{}) {
```
brandonbloom, line 25: this isn't done at all :-P

brandonbloom:
Switched reviewer to @BenElgar, since Andrew and I worked on this together and could really use another set of eyes / more testing before merging/releasing.

BenElgar (commented):
This seems to break volumes. If I apply the following docker-compose file:
```
services:
  t0:
    image: bash
    command: sleep infinity
    volumes:
      - db-data:/etc/data
volumes:
  db-data: {}
```
I get the following error:
```
⨯ └─ updating t0 delete "t0" for replacement: unmarshalling spec: interpolating: at [volumes 0 volume]: cannot interpolate value of type: <nil>
```

BenElgar:
This feature also interfaces badly with this bug: https://github.com/deref/exo/issues/270. Because it's easy to omit an env var you can easily get yourself into a state where you have to nuke the workspace and start again — as I did many times whilst testing.

brandonbloom:
I've force-pushed over this branch with a more-up-to-date version of the code that was here before, but reverted this back to a draft because I've found numerous additional problems while working on it.

brandonbloom:
Also worth noting that many small bits of this have made it to main via other PRs.

brandonbloom:
OK, this appears to be in working condition again, but I am hesitant to merge it as there are likely to be many bugs still. Going to sit on it for a little bit & work on more robustness things tomorrow.

brandonbloom:
This branch is now super stale with respect to #418

PR #250, opened 2021-08-25
New component & container pages

jwmza:
This adds a page to the GUI for creating a new container
image

as well as:

- Adds a **New Component** screen to select which type of component you'd like to add to a workspace
- Cleans up existing creation forms and updates the overall design&layout
- Trims `:out` and `:err` from logs in the GUI, similar to the command line log output.

Issue #208, opened 2021-08-17
[BUG] process list UX still confusing


brandonbloom:
![image](https://user-images.githubusercontent.com/119164/129799930-b97a796b-0dcb-47ab-89b6-dfab2ada2b34.png)

brandonbloom:
From a user in dark mode:

> One small confusion with the UI is the "Stop process" button, instead of a grey square maybe change it to red or something since the grey one looks like I'm selecting the process.

jwmza:
Is this complete with the new start/stop icons?

![image](https://user-images.githubusercontent.com/51100181/132732637-7795aa67-d138-4b4d-9748-1766393c7095.png)
![image](https://user-images.githubusercontent.com/51100181/132732660-5dc11fc4-9624-4532-b26e-195ab3de5ec9.png)

jwmza:
Should be noted that this UI will materially change when we remake this list as a cards list.

I also still think the running/not running indicator is not ideal, plus we need some sort of broken/errored icon for components that are incorrectly configured or can't run.

jwmza:
Okay I think this "bug" is fixed now - there will always be improvements to be made in future, but the new cards list is good in my books. Feel free to re-open or make a new issue regarding a particular problem with it if you disagree :)

PR #267, opened 2021-08-30
Improve Process List styling

jwmza:
This is a quick fix towards #208 

image
image

Not the ideal solution yet - there is still room for improvement in this process control list with the introduction of better tooltips (as we now have in the navbar, but we might need a good way to get tooltips overlaid in-position but outside of a parent who has hidden `overflow`. It also will be necessary soon to replace the Delete button in the process list with an ellipsis dropdown menu containing actions like delete and modify, but this will be introduced after #250 , which will introduce both some UI components necessary for these kind of dropdowns as well as the process modification page.


jwmza:
- Uses solid instead of line icons for process run controls: with a two-rectangle "pause" type icon for stopping a process. Perhaps we should use something like a red square instead; and a solid pay icon for starting a process.
- Changes the dull gray square icon for a passive not-running process into an empty outlined circle the same size as the green running process indicator icon.

brandonbloom:
new screenshots please :)

brandonbloom (approved):
LGTM

PR #273, opened 2021-09-01
Wait for log forwarding to finish before exiting

BenElgar:


internal/supervise/main.go
```diff
@@ -126,11 +127,25 @@ func Main() {
 
 	// Proxy logs.
 	syslogProcID := strconv.Itoa(child.Pid)
-	go pipeToSyslog(ctx, conn, cfg.ComponentID, "out", syslogProcID, stdout)
-	go pipeToSyslog(ctx, conn, cfg.ComponentID, "err", syslogProcID, stderr)
 
-	// Wait for child process to exit.
+	var wg sync.WaitGroup
+	work := func(f func()) {
```
brandonbloom, line 132: instead of introducing this `work` function, you can just do `wg.Add(2)` and leave the `go pipeToSyslog` calls alone.

brandonbloom: oh er nevermind, ignore me here

brandonbloom (approved):

PR #276, opened 2021-09-01
Implement staged recreate for apply

### LINKED ISSUE #275: [BUG] When applying a docker-compose file, the workspace can get in a bad state
**Describe the bug**
When running `exo apply` in a workspace with a Docker Compose file, the components will not be recreated in the correct order, so we can end up with components referencing Docker resources that no longer exist, unmet dependencies, and orphaned Docker resources. There are several different manifestations of bugs that are all related to this behavior.

**To Reproduce**
1. Enter a workspace with a docker-compose.yml file.
2. Run `exo apply`
3. Re-run `exo apply`

**Expected behavior**
The components defined in `docker-compose.yml` are deleted in reverse dependency order then recreated in dependency order.

kendru:
This PR changes the order in which components are recreated such that we never have a component that references a nonexistent resource.

Fixes #275

### DIFF
--- a/internal/core/server/workspace.go
+++ b/internal/core/server/workspace.go
@@ -113,54 +113,85 @@ func (ws *Workspace) Apply(ctx context.Context, input *api.ApplyInput) (*api.App
 		return nil, fmt.Errorf("describing components: %w", err)
 	}
 
-	// Index old components by name.
 	oldComponents := make(map[string]api.ComponentDescription, len(describeOutput.Components))
-	for _, component := range describeOutput.Components {
-		oldComponents[component.Name] = component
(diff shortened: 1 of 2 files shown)


brandonbloom (approved):
I _think_ this LGTM, but to be totally honest, I haven't invested sufficient time to wrap my head around the algorithm. The code looks all reasonable though.

kendru:
@brandonbloom Thanks for looking it over. It is definitely too complicated, and I'm not too happy with it, but it does solve a pretty nasty bug, so I'll go ahead and merge it, then we can decide if/how to better handle `apply`.

PR #277, opened 2021-09-01
Parse extra fields in compose file

### LINKED ISSUE #227: [BUG] incomplete yaml support: anchors, aliases, and merge keys
See the YAML spec https://yaml.org/spec/1.2/spec.html

This blog post shows an example of using these features for compose files:
https://medium.com/@kinghuang/docker-compose-anchors-aliases-extensions-a1e4105d70bd

kendru:
Fixes #227 

- Let YAML parser parse all fields into an extra `Raw` map.
- Manually validate keys, and throw an error if a nonstandard, non-extension field is found.

This allows us to support the pattern of using an extension field to define common config. See https://medium.com/@kinghuang/docker-compose-anchors-aliases-extensions-a1e4105d70bd for an example.

### DIFF
--- a/internal/providers/docker/compose/project.go
+++ b/internal/providers/docker/compose/project.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"io"
 	"strconv"
+	"strings"
 
 	"code.cloudfoundry.org/bytefmt"
 	"github.com/goccy/go-yaml"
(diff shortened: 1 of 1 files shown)


internal/providers/docker/compose/project.go
```diff
@@ -20,12 +21,24 @@ import (
...
 		return nil, err
 	}
+
+	// Validate.
+	for key := range comp.Raw {
+		switch key {
+		case "version", "services", "networks", "volumes", "configs", "secrets":
+			// Ok.
+		default:
+			if !strings.HasPrefix(key, "x-") {
```
BenElgar, line 36: What's the `x-` prefix?

internal/providers/docker/compose/project.go
```diff
@@ -20,12 +21,24 @@ import (
...
+				return nil, fmt.Errorf("unsupported top-level key in compose file: %q", key)
```
brandonbloom, line 37: What about doing this recursively for non-top-level keys?

BenElgar (approved):
Happy with this though I'd ask for a comment explaining why we don't disallow unknown fields.

internal/providers/docker/compose/project.go
brandonbloom: https://github.com/compose-spec/compose-spec/blob/master/spec.md#extension

internal/providers/docker/compose/project.go
kendru: Yeah, it looks like compose 3.4 only specified extension fields at the top level, but the new compose spec allows them to appear as the keys of any object. Since that is the case, I'll do this recursively.

brandonbloom:
Maybe instead of Ben's suggestion of a comment, we should put an example with anchors in it in a test :)

PR #278, opened 2021-09-01
Add support for arbitrary text as shell command

BenElgar:


internal/providers/docker/components/container/image.go
```diff
@@ -34,7 +34,10 @@ func (c *Container) ensureImage(ctx context.Context) error {
 	c.State.Image.Command = inspection.Config.Cmd
 	c.State.Image.WorkingDir = inspection.Config.WorkingDir
 	c.State.Image.Entrypoint = inspection.Config.Entrypoint
-
+	c.State.Image.Shell = inspection.Config.Shell
+	if len(c.State.Image.Shell) == 0 {
+		c.State.Image.Shell = []string{"/bin/sh", "-c"}
```
brandonbloom, line 39: Looking at the code for docker engine and I see this:

```go
func defaultShellForOS(os string) []string {
	if os == "linux" {
		return []string{"/bin/sh", "-c"}
	}
	return []string{"cmd", "/S", "/C"}
}
```

I believe the OS is available in `inspection.Os`

internal/providers/docker/compose/command.go
```diff
@@ -1,34 +1,41 @@
 package compose
 
-import (
-	"github.com/mattn/go-shellwords"
-)
-
 // FIXME: this incorrectly treats the "command" passed to docker as a meaningful
```
brandonbloom, line 3: old comment now? maybe should document the `IsShellForm` boolean instead.

internal/providers/docker/components/container/image.go
BenElgar: Made the change but I'm not a big fan of adding code I can't meaningfully test.

PR #280, opened 2021-09-01
Add commands to get/set/clear/edit component state

### LINKED ISSUE #271: [FEATURE] state repair CLI commands

Sometimes, things go wrong with component state. See #270 for example. While we should clearly fix that bug, I've still had to do state surgery in every IaC tool I've ever used. We should have some basic tools to assist with this.

I'm thinking a family of new `state` commands:

```
exo state get COMPONENT
exo state set COMPONENT < whatever.json
exo state clear COMPONENT
```

I'd probably mark the `state` subcommand group as "hidden", since _generally_ the `refresh` command should be all you need to repair unless something has gone wrong. However, when implementing providers, something _frequently_ goes wrong.

**Describe alternatives you've considered**

Hand editing the state file. This won't be possible once we switch to SQLite for storage.

kendru:
Adds the following commands:

- `exo state  get COMPONENT` - dumps component state to stdout
- `exo state set COMPONENT` - reads JSON from stdin and saves it as new component state
- `exo state clear COMPONENT` - sets component state to be an empty object
- `exo state edit COMPONENT` - opens component state in system editor and saves changes

Fixes #271

### DIFF
--- a/cmd/exo/state.go
+++ b/cmd/exo/state.go
@@ -1,0 +1,192 @@
+package main
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/deref/exo/internal/core/api"
+	"github.com/deref/exo/internal/util/jsonutil"
+	"github.com/deref/exo/internal/util/which"
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	rootCmd.AddCommand(stateCmd)
+	stateCmd.AddCommand(stateGetCmd)
+	stateCmd.AddCommand(stateSetCmd)
+	stateCmd.AddCommand(stateClearCmd)
+	stateCmd.AddCommand(stateEditCmd)
+
+	stateCmd.AddCommand(helpSubcmd)
+}
+
+var stateCmd = &cobra.Command{
+	Use:    "state",
+	Short:  "View and update the state store.",
+	Long:   `Contains subcommands for getting, setting, and clearing state on a per-component basis.`,
+	Hidden: true,
+	Args:   cobra.MaximumNArgs(0),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return nil
+	},
+}
+
+var stateGetCmd = &cobra.Command{
+	Use:    "get <component>",
+	Short:  "Print component state",
+	Hidden: true,
+	Args:   cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		componentRef := args[0]
+		ctx := newContext()
+		checkOrEnsureServer()
+		cl := newClient()
+
+		workspace := requireWorkspace(ctx, cl)
+		output, err := workspace.GetComponentState(ctx, &api.GetComponentStateInput{Ref: componentRef})
+		if err != nil {
+			return err
+		}
+
+		return jsonutil.PrettyPrintJSONString(os.Stdout, output.State)
+	},
+}
+
+var stateSetCmd = &cobra.Command{
+	Use:    "set <component>",
+	Short:  "Set component state",
+	Long:   "Set component state to the JSON received on stdin.",
+	Hidden: true,
+	Args:   cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		componentRef := args[0]
+		ctx := newContext()
+		checkOrEnsureServer()
+		cl := newClient()
+
+		var newState map[string]interface{}
+		if err := json.NewDecoder(os.Stdin).Decode(&newState); err != nil {
+			return fmt.Errorf("reading state from stdin: %v", err)
+		}
+
+		workspace := requireWorkspace(ctx, cl)
+		_, err := workspace.SetComponentState(ctx, &api.SetComponentStateInput{
+			Ref:   componentRef,
+			State: jsonutil.MustMarshalString(newState),
+		})
+		return err
+	},
+}
+
+var stateClearCmd = &cobra.Command{
+	Use:    "clear <component>",
+	Short:  "Clear component state",
+	Hidden: true,
+	Args:   cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		componentRef := args[0]
+		ctx := newContext()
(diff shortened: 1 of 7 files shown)


brandonbloom:
<3 you for doing an edit command :)

Will look at the code tomorrow. Or @BenElgar can maybe review.

cmd/exo/state.go
```diff
@@ -0,0 +1,192 @@
...
+		defer os.Remove(tmpfile.Name())
+
+		if err := jsonutil.PrettyPrintJSONString(tmpfile, output.State); err != nil {
+			return fmt.Errorf("writing to temporary file: %w", err)
+		}
+		if err := tmpfile.Close(); err != nil {
+			return fmt.Errorf("closing temporary file: %w", err)
+		}
+
+		// TODO: Read from EDITOR, or guess from common editors.
```
BenElgar, line 137: Is this not doing that?

cmd/exo/state.go
```diff
@@ -0,0 +1,192 @@
...
+	stateCmd.AddCommand(helpSubcmd)
+}
+
+var stateCmd = &cobra.Command{
+	Use:    "state",
+	Short:  "View and update the state store.",
+	Long:   `Contains subcommands for getting, setting, and clearing state on a per-component basis.`,
+	Hidden: true,
+	Args:   cobra.MaximumNArgs(0),
+	RunE: func(cmd *cobra.Command, args []string) error {
```
BenElgar, line 34: Shouldn't we just omit `RunE`?

cmd/exo/state.go
```diff
@@ -0,0 +1,192 @@
...
+
+	stateCmd.AddCommand(helpSubcmd)
+}
+
+var stateCmd = &cobra.Command{
+	Use:    "state",
+	Short:  "View and update the state store.",
+	Long:   `Contains subcommands for getting, setting, and clearing state on a per-component basis.`,
+	Hidden: true,
+	Args:   cobra.MaximumNArgs(0),
```
BenElgar, line 33:
```suggestion
	Args:   cobra.NoArgs,
```
I think this returns a slightly clearer error.

cmd/exo/state.go
```diff
@@ -0,0 +1,192 @@
...
+
+		edit := exec.Command(editor, tmpfile.Name())
+		edit.Stdin = os.Stdin
+		edit.Stdout = os.Stdout
+		edit.Stderr = os.Stderr
+		if err := edit.Run(); err != nil {
+			return fmt.Errorf("editing state file: %w", err)
+		}
+
+		newState, err := os.ReadFile(tmpfile.Name())
```
BenElgar, line 181: Minor but I'd argue that we should only update the state if the temporary file was saved. Otherwise we leave people no way of exiting without doing an update.

In an ideal world, the temp file would act as a two way binding to the component's state.

BenElgar (approved):

cmd/exo/state.go
```diff
@@ -0,0 +1,192 @@
...
+	Args:   cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		componentRef := args[0]
+		ctx := newContext()
+		checkOrEnsureServer()
+		cl := newClient()
+
+		var newState map[string]interface{}
+		if err := json.NewDecoder(os.Stdin).Decode(&newState); err != nil {
+			return fmt.Errorf("reading state from stdin: %v", err)
```
brandonbloom, line 74:
```suggestion
			return fmt.Errorf("reading state from stdin: %w", err)
```

cmd/exo/state.go
```diff
@@ -0,0 +1,192 @@
...
+
+			for _, candidateEditor := range []string{
+				"code",
+				"vim",
+				"nano",
+				"vi",
+				"emacs",
+				"ee",
+			} {
+				found, err := which.Query{
```
brandonbloom, line 154: I could have sworn I had made a global `which.Which` function that used the current working directory, PATH variable, etc. Would be a nice small cleanup here.

cmd/exo/state.go
```diff
@@ -0,0 +1,192 @@
...
+			}
+			pathVar, _ := os.LookupEnv("PATH")
+
+			for _, candidateEditor := range []string{
+				"code",
+				"vim",
+				"nano",
+				"vi",
+				"emacs",
+				"ee",
```
brandonbloom, line 152: ... I've never heard of `ee`.... what is this?

cmd/exo/state.go
```diff
@@ -0,0 +1,192 @@
...
+		// TODO: Read from EDITOR, or guess from common editors.
+		editor := os.Getenv("EDITOR")
+		if editor == "" {
+			wd, err := os.Getwd()
+			if err != nil {
+				return fmt.Errorf("getting working directory: %w", err)
+			}
+			pathVar, _ := os.LookupEnv("PATH")
+
+			for _, candidateEditor := range []string{
```
brandonbloom, line 146: What does git or whatever do? Is this a reasonable approach to just guess an editor? I guess it seems fine ¯\_(ツ)_/¯

brandonbloom (approved):
Still, LGTM with minor comments. Just had 10 min between meetings to give it a closer look :)

cmd/exo/state.go
kendru: Oops - forgot to remove the TODO comment!

brandonbloom: I know that all three of us have done that about a half dozen times each over the last year :)

cmd/exo/state.go
kendru: Good call. I'll update the code to do this.

cmd/exo/state.go
kendru: I didn't see one, but I just refactored this into a new `which.Which`

cmd/exo/state.go
kendru: It's the default editor on FreeBSD, and I think some other Unix-like systems.

cmd/exo/state.go
kendru: It looks like it uses EDITOR or a git-configured editor, then (at least on Debian-bases systems) it falls back to `editor` or `sensible-editor`, which are symlinks to a real editor. I'll add those to the beginning of the search list.

PR #283, opened 2021-09-02
More compose features

kendru:
This is my pass at getting closer to feature complete with the settings in a Compose file.


brandonbloom:
@kendru I tried to set the target on this to the component-state-tools branch, but seems like there was some rebasing going on. i'll leave you alone until this is a non-draft :-P

internal/providers/docker/components/container/lifecycle.go
```diff
@@ -421,3 +470,129 @@ func (c *Container) removeExistingContainerByName(ctx context.Context) error {
 
 	return nil
 }
+
+func (c *Container) endpointSettings(sn compose.ServiceNetwork) *network.EndpointSettings {
+	es := &network.EndpointSettings{
+		Aliases: append([]string{c.ComponentID, c.ComponentName}, sn.Aliases...),
```
brandonbloom, line 476: Does docker-compose do something similar here or is this some aliasing logic we're adding for convenience?

What happens if there is a conflict on these aliases?

internal/providers/docker/compose/project_test.go
```diff
@@ -173,10 +598,24 @@ func TestParseService(t *testing.T) {
...
+				return
+			}
 
 			svc := comp.Services["test-svc"]
 			assert.Equal(t, expected, svc)
 		})
 	}
 }
+
+func boolRef(b bool) *bool {
```
brandonbloom, line 611: I've seen functions like these in so many projects and it's such a bummer that they are necessary :-P

brandonbloom (approved):
LGTM, but coordinate with @BenElgar b/c iirc there is a potential conflict with the env file work?

internal/providers/docker/components/container/lifecycle.go
kendru: Yes, this is essentially what docker-compose does as well, and we will use the same names as long as we have not mangled the name to fit the component naming convention. Since we find and replace any containers that Docker Compose created, there should not be any conflict.

internal/providers/docker/compose/project_test.go
kendru: Yep! It would be really nice if there was a generic `HeapAlloc(T) -> *T` function :D

kendru:
> LGTM, but coordinate with @BenElgar b/c iirc there is a potential conflict with the env file work?

I resolved the conflicts with his changes, so this should incorporate his changes as well.

PR #288, opened 2021-09-03
Add support for numbers as env vars

BenElgar:


internal/providers/docker/compose/dictionary.go
```diff
@@ -24,12 +47,12 @@ func (dict *Dictionary) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
 	switch data := data.(type) {
 	case map[string]interface{}:
-		for k, v := range data {
-			var ok bool
-			s, ok := v.(string)
-			if !ok {
-				return fmt.Errorf("expected values to be string, got %v", v)
-			}
+		stringOrNumMap := make(map[string]stringOrNum)
+		if err := unmarshal(&stringOrNumMap); err != nil {
```
brandonbloom, line 51: Why do an extra remarshal pass? Maybe I'm missing something, but it seems like you could do a `switch v := v.(type)` and have cases for string, and the various number types.

BenElgar: Yes, but it's my understanding that we'd need to convert this into a bye unmarshaller rather than an interface unmarshaller, since we need the original string, and then add cases for around ten[0] number types. That seemed less clear than this.

[0] https://golang.org/ref/spec#Numeric_types

PR #289, opened 2021-09-03
Add support for env files

BenElgar:


internal/providers/docker/components/container/lifecycle.go
```diff
@@ -62,6 +64,18 @@ func (c *Container) create(ctx context.Context) error {
 	}
 
 	envMap := map[string]string{}
+	for _, envFilePath := range c.Spec.EnvironmentFiles {
+		if !path.IsAbs(envFilePath) {
```
brandonbloom, line 68: After resolving the path, please check `pathutil.HasPathPrefix` - see the `resolveWorkspacePath` function. As a potential future cleanup, it would be nice if we could somehow pass in a pre-scoped filesystem (fs.FS, i believe) interface for components to make it easier to keep this secure by default.

internal/providers/docker/components/container/lifecycle.go
```diff
@@ -62,6 +64,18 @@ func (c *Container) create(ctx context.Context) error {
...
+			envFilePath = path.Join(c.WorkspaceRoot, envFilePath)
+		}
+		envFileVars, err := godotenv.Read(envFilePath)
+		if err != nil {
+			return fmt.Errorf("reading env file %s: %w", envFilePath, err)
```
brandonbloom, line 73:
```suggestion
			return fmt.Errorf("reading env file %q: %w", envFilePath, err)
```

brandonbloom (approved):

kendru:
@BenElgar it looks like this might conflict with #283, but I have not implemented actually resolving the environment files, so I'll probably just resolve in favor of this.

internal/providers/docker/components/container/lifecycle.go
BenElgar: Do we want to do this? The spec implies that environment files can be located anywhere on the host. Happy to deviate if you think we should though.

https://github.com/compose-spec/compose-spec/blob/master/spec.md#env_file

brandonbloom: Yes, I think we want to do this.

Enforcing this restriction strongly encouraging reproducible, encapsulated, version-controlled environment configurations. Another benefit is avoiding potential security issues, if/when we ever run any exo components in the cloud.

I understand that it is an incompatibility. However, it's only a _little_ incompatible with the spec, since absolute paths would still work, but they just have to be within the workspace root. It seems pretty unlikely that any real project in practice will use a path outside of the project root, except if some user overrides an environment variable to be some global path on their system.

Also, let's remember that we can always relax this constraint, but it would be difficult to tighten it again.

PR #293, opened 2021-09-03
command to kill/signal processes

brandonbloom:
implements #188

### DIFF
--- a/cmd/exo/kill.go
+++ b/cmd/exo/kill.go
@@ -1,0 +1,47 @@
+package main
+
+import (
+	"github.com/deref/exo/internal/core/api"
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	rootCmd.AddCommand(killCmd)
+	killCmd.Flags().StringVarP(&killFlags.Signal, "signal", "s", "kill", "")
+}
+
+var killFlags struct {
+	Signal string
+}
+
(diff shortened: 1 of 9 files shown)


brandonbloom:
failure appears to be a flakey test, will fix independently

cmd/exo/kill.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+var killCmd = &cobra.Command{
+	Use:   "kill [ref...]",
+	Short: "Force stop processes",
+	Long: `Sends a unix signal to process-like components.
+	
+If no refs are provided, signals all processes in the workspace.
+	
+The default signal is SIGKILL.`,
```
BenElgar, line 24: SIGKILL being the default feels strange when other kill commands generally use SIGTERM as the default.

BenElgar (commented):
LGTM aside from small comment

cmd/exo/kill.go
brandonbloom: Understood, but 1) `stop` already does sigterm, then timeout, then sigkill. 2) 90% of the time when I use unix `kill`, I use `kill -9` and 3) this is the default in docker compose.

BenElgar: :+1: Persuaded

PR #295, opened 2021-09-04
add variables gui

### LINKED ISSUE #272: [FEATURE] environment variables CLI/GUI

I want to see what the effective environment variables are in my workspace. In the future, I'd like to also be able to add/remove/edit variables too.

Command to print the current workspace environment as a .env file:

```
exo env
```

And a "Variables" page somewhere in the GUI showing the same information.

brandonbloom:
Closes #272

image


brandonbloom:
Hey @jwmza, this is like 90% done, but could use a nice icon. I was thinking something like `$x` for "variables". Maybe also a quick fix or tweak on table spacing or whatnot.

Would you please carry it across the finish line for me? Thanks!

brandonbloom:
To elaborate on the table "spacing" issue, the problem is that very long environment variable names cause the table to grow and fall off the right side of the GUI. Need to add some overflow settings or something.

jwmza (commented):
Current code LGTM, I'll add an icon and fix the overflow issues now

jwmza:
Icon:
image

jwmza:
@brandonbloom the solution that was already in place (in the old built-in table for `Process.svelte`) was not ideal, as it clamps to 600px instead of filling the available space while also overflowing, or doing something more intelligent like that - but it does *work*. Unfortunately it is strangely difficult to properly size such things unless you do an arbitrary rigid layout with a grid or something like that, and the reason is because HTML/CSS were not made for this kind of modern GUI application and the geometry is still very much informed by what the languages were once used for. I'll be continuously working to improve such sizing and overflow issues.

Anyways, the reason it started overflowing again is because when the environment table was extracted to its own component, the CSS at the bottom of the file relevant to the table was missed. This is probably because there's no warning system or errors for such partial component extractions, so it seems the only way to prevent this is to be aware of exactly which styles need to be cut and pasted or copy and pasted to the resulting file after extraction. Some styles only apply to the component you're extracting, and some apply to both the component and other elements in the parent page, so it's a bit of a hassle.

See [this commit](https://github.com/deref/exo/pull/295/commits/cd823d53376fb279739b7d41e3abba6dc4f0c685) for what I mean.

gui/src/pages/Process.svelte
```diff
@@ -81,37 +86,47 @@
...
-                  /></td
-                >
+                  />
+                </td>
               </tr>
               <tr>
                 <td class="label">Resident Memory</td>
-                <td><BytesLabel value={process.residentMemory} /></td>
+                <td>
+                  {#if process.residentMemory}
```
jwmza, line 107: I added some conditional wrappers to fix all the typescript errors this file was causing, but perhaps this is not the ideal way to handle these errors and we could fix the problem upstream where `process` is defined. What do you think @brandonbloom ?

gui/src/pages/WorkspaceVariables.svelte
```diff
@@ -0,0 +1,31 @@
...
+  import Spinner from '../components/Spinner.svelte';
+  import WorkspaceNav from '../components/WorkspaceNav.svelte';
+  import EnvironmentTable from '../components/EnvironmentTable.svelte';
+  import { api } from '../lib/api';
+
+  export let params = { workspace: '' };
+
+  const workspaceId = params.workspace;
+  const workspace = api.workspace(workspaceId);
+  const workspaceRoute = `/workspaces/${encodeURIComponent(workspaceId)}`;
```
jwmza, line 13: These 3 definitions are ALL OVER the app. Almost every page has a duplication of this code. I'm not sure the best way to apply DRY here but I'm thinking maybe we should extract this to a library function that is called in a single line like

```
const { workspaceId, workspace, workspaceRoute } = someFunction(params)
```

jwmza (approved):
Take a look at these comments please but otherwise, this looks good to me.

gui/src/pages/WorkspaceVariables.svelte
jwmza: ^ (If this is improved it probably makes sense to do it outside of this PR)

gui/src/pages/Process.svelte
```diff
@@ -47,10 +48,12 @@
         process = processList.data.filter((p) => p.id === processId)[0];
       }
       if (process && process.running) {
-        cpuPercentages.push(process.cpuPercent);
+        process.cpuPercent && cpuPercentages.push(process.cpuPercent);
```
brandonbloom, line 51: @jwmza Is this value potentially zero? Should this be `cpuPercent != null`?

gui/src/pages/Process.svelte
brandonbloom: What were the errors before? That process may have fields with null or undefined values? Seems reasonable to handle those conditionally.

gui/src/pages/WorkspaceVariables.svelte
brandonbloom: Yeah, definitely for another PR. There are a variety of ways to DRY this up we can look at as we continue to add pages.

brandonbloom:
Thanks @jwmza!

PR #297, opened 2021-09-06
Use the same build context as docker build command

BenElgar:
This is an attempt to ensure that we always build the same image that `docker build` would have (assuming BuildKit is not being used).

### DIFF
--- a/internal/providers/docker/components/container/build.go
+++ b/internal/providers/docker/components/container/build.go
@@ -54,6 +54,7 @@ func (c *Container) buildImage(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("getting build context: %w", err)
 	}
+	defer buildContext.Close()
 
 	opts := types.ImageBuildOptions{
 		//Tags           []string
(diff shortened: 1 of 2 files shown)


internal/providers/docker/components/container/build.go
```diff
@@ -37,139 +35,134 @@ func (c *Container) canBuild() bool {
...
-	})
+	contextPath := filepath.Join(c.WorkspaceRoot, spec.Build.Context)
+	if !pathutil.HasFilePathPrefix(contextPath, c.WorkspaceRoot) {
+		return errors.New("docker container build context path must be in exo workspace root")
+	}
+	dockerfile := c.Spec.Build.Dockerfile
+	if dockerfile == "" {
+		dockerfile = "Dockerfile"
+	}
+	buildContext, err := getArchive(contextPath, dockerfile)
```
brandonbloom, line 53: This takes ownership of the reader, so you need to do a `defer buildContext.Close()`. If you don't things seem like they will be fine on the happy path, but if an error occurs reading from the archive, then the writing goroutine inside of `archive.TarWithOptions` will block indefinitely & this will be a goroutine leak.

brandonbloom (approved):
LGTM except one goroutine leak on error.

brandonbloom:
OK then, let's close this when #297 is merged & we can re-open if it ever becomes an issue. Thanks!

internal/providers/docker/components/container/build.go
BenElgar: Good catch!

PR #303, opened 2021-09-07
Add manifest support

BenElgar:
The idea here is to extract the manifest conversion from the manifest loading. The manifest will first be converted, the converted version saved and then the converted one actually applied.

The major thing missing from this right now is the writing back of manifests when things are changed in the UI but the infrastructure is there to preserve comments and formatting.

**procfile**
```Procfile
t0: /home/ben/Programming/exo/bin/tick
```
```yaml
exo: "0.1"
components:
  - name: t0
    type: process
    spec:
        program: /home/ben/Programming/exo/bin/tick
        environment:
            PORT: "5000"
```

**docker-compose**
```yaml
services:
  t0:
    image: bash
    command: "sleep infinity"
    env_file: my-env
```
```yaml
exo: "0.1"
components:
  - name: default
    type: network
    spec:
        driver: bridge
        name: tmpMQvlHcwsZ4_default
  - name: t0
    type: container
    spec:
        command: sleep infinity
        container_name: tmpMQvlHcwsZ4_t0_1
        env_file: my-env
        image: bash
        labels:
            com.docker.compose.project: tmpMQvlHcwsZ4
            com.docker.compose.service: t0
        networks:
            tmpMQvlHcwsZ4_default:
                network: tmpMQvlHcwsZ4_default
    depends_on:
      - default
```


internal/providers/unix/components/process/controller.go
```diff
@@ -6,10 +6,11 @@ import (
 	"fmt"
 
 	"github.com/deref/exo/internal/util/jsonutil"
+	"github.com/goccy/go-yaml"
 )
 
 func (p *Process) InitResource() error {
-	if err := jsonutil.UnmarshalString(p.ComponentSpec, &p.Spec); err != nil {
+	if err := yaml.Unmarshal([]byte(p.ComponentSpec), &p.Spec); err != nil {
```
brandonbloom, line 13: 1) why change this?
2) this will be a breaking change for any existing users

BenElgar: 1. To make it consistent with the docker-compose version.
2. I wouldn't think it would be a breaking change given that YAML is superset of JSON.

brandonbloom: 1. not saying that it shouldn't be consistent, just wonder why it should be?
2. wow.... i had no idea that was true...

cmd/exo/apply.go
```diff
@@ -58,23 +59,52 @@ var applyCmd = &cobra.Command{
...
+		if !path.IsAbs(manifestPath) {
+			manifestPath = path.Join(workingDir, manifestPath)
 		}
-		s := string(bs)
-		input.Manifest = &s
 	}
-	if applyFlags.Format != "" {
-		input.Format = &applyFlags.Format
+
+	for _, candidate := range manifestCandidates {
```
brandonbloom, line 75: Why is this search part of the CLI? Seems like as much of this logic as possible should be part of the server, so that the GUI can reuse the logic.

cmd/exo/convert.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+var convertFlags struct {
+	Format string
+}
+
+var convertCmd = &cobra.Command{
+	Use:   "convert [flags] [manifest-file]",
+	Short: "Converts a docker-compose file or Procfile into an exo manifest",
+	Long: `If unspecified, a manifest format will be guessed from the manifest filename.  This can be
```
brandonbloom, line 31: iirc, the short message doesn't appear at all when you do `exo help convert`, and so you should more-or-less duplicate the short string in to the long string as well.

internal/providers/unix/components/process/controller.go
brandonbloom: ah, ok - i think i see your answer to #1 now: so you can do consistent handling with the load/save/etc without having to switch-on-type or soemthing like that. works for me.

cmd/exo/apply.go
BenElgar: Yes I think there's follow up work to create a `get-manifest-candidates` command on the server that returns a list of possible manifests in priority order.

brandonbloom: It's not obvious from looking at the changes: Why does the CLI need this at all?

BenElgar: Given the assumption that everyone will have an exo file (which is perhaps an unfinished discussion), I think it makes sense that the `apply` endpoint takes just one format of exo file and that its responsibility is to simply apply it, not to possibly convert it, save it  and then apply a converted manifest. In order to achieve that, I moved the conversion and saving steps into the CLI. I'm also fine with having a `convert` endpoint that does the conversion.

The actual locating of the file feels like a client concern: for the CLI running a manifest in the working directory seems like a reasonable choice but for the UI you might want to have a `run` button which gives you a choice between manifests. Something like this:

brandonbloom: > Given the assumption that everyone will have an exo file

let's discuss more via slack

> The actual locating of the file feels like a client concern

I think that some change in perspective is necessary here. The word "client" is sort of ambiguous/meaningless in the context of exo. The CLI tool is a user interface. In terms of separation of concerns, it should only be concerned with command parsing, terminal rendering, etc. From this perspective, searching for manifest files is a _workspace_ concern, and therefore should be centralized where any UI (ie the CLI or GUI) can access it.

> a run button which gives you a choice between manifests

I think about it as having a choice of run configurations - probably called stacks - instead of a choice of manifests. Seems like stack = manifest + config.

brandonbloom: back to the manifest search logic:

One bit of feedback we've received is that somebody didn't realize they had to run `apply` after their manifest changed. We're also planning on adding a file watcher anyway, so it would be good that if a file changes, the server knows if it is or could potentially be a manifest file and some UI shows that there is a diff between the state and the manifest. Then you'd want to either commit the change to the manifest, or apply the manifest to update the state.

internal/providers/docker/compose/servicedependency.go
```diff
@@ -54,16 +52,15 @@ func (sd *ServiceDependencies) UnmarshalYAML(unmarshal func(interface{}) error)
 		return nil
 	}
 
-	var asMap yaml.MapSlice
+	var asMap map[string]interface{}
```
brandonbloom, line 55: This introduces various bugs in the form of instability. Go maps are intentionally unordered keys. If you're serializing to a map, the encoder might sort the keys, but that's not the right thing here. We want to preserve the key order when we convert to a slice below (and in several other places throughout the code where you changed this).

BenElgar: Why? I was treating these arrays as sets rather than ordered lists. Is the order pertinent for anything other than serialisation? My intent was to sort them at the point of serialisation to preserve the order of the original map.

cmd/exo/apply.go
BenElgar: > searching for manifest files is a workspace concern

Yes, but figuring out what to do with them once you've found them feels like a UI concern. In the CLI you probably want to prefer the one in your working directory. In the GUI you probably want to choose the one in the workspace root. So I'd propose a `get-manifest-candidates` which returns all possible manifests and then the GUI or the CLI makes the final choice about what to do.

BenElgar:
Reverted merging master since it's currently broken for me.

internal/providers/unix/components/process/controller.go
brandonbloom: btw, I think there is a yamlutil.UnmarshalString too

internal/util/yamlutil/render.go
```diff
@@ -0,0 +1,39 @@
+package yamlutil
+
+import (
+	yaml "gopkg.in/yaml.v3"
+)
+
+// MergeFormatting merges the formatting of the original yaml into the updated
+// version. At the moment it's not tremendously sophisticated and just preserves
+// comments.
+func MergeFormatting(original *yaml.Node, updated *yaml.Node) error {
```
brandonbloom, line 10: This probably needs some tests :)

cmd/exo/apply.go
brandonbloom: Fair point, I had not considered working directory vs workspace root. Though it's not yet clear to me how that should work, especially if we get in to hierarchical components for subsystems and the like.

I'm not opposed to the `get-manifest-candidates` API in general, my concern is the remaining extra logic now of the conversion, etc happening in the CLI. Can we do some design work on the UI so we know how this is going to work there? I want to make sure we have really strong CLI/GUI parity & I suspect that this will need to be refactored back closer to the original separation of concerns pretty much immediately upon trying to add an "apply" feature to the UI.

brandonbloom:
Going to close this because it's pretty far out of date now & I'm on a massive refactor journey that will eventually culminate in a new manifest format.

PR #304, opened 2021-09-08
Add NavbarRoute component for repetitive route NavbarButtons

### LINKED ISSUE #251: Clean up `NavbarButton` click handlers
Maybe `NavbarButton` should have a `href` property, which would clean up all these `on:click` handlers.

_Originally posted by @brandonbloom in https://github.com/deref/exo/pull/241#discussion_r695928130_

jwmza:
Fix #251

### DIFF
--- a/gui/src/components/nav/NavbarRoute.svelte
+++ b/gui/src/components/nav/NavbarRoute.svelte
@@ -1,0 +1,18 @@
+<script lang="ts">
+  import NavbarButton from './NavbarButton.svelte';
+  import * as router from 'svelte-spa-router';
+
+  export let title: string | undefined = undefined;
+  export let active: string | undefined = undefined;
+  export let href: string;
+</script>
+
+<NavbarButton
+  {title}
+  {active}
+  on:click={() => {
+    router.push(href);
+  }}
+>
+  <slot />
+</NavbarButton>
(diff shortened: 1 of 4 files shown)


gui/src/components/nav/NavbarRoute.svelte
```diff
@@ -0,0 +1,18 @@
+<script lang="ts">
+  import NavbarButton from './NavbarButton.svelte';
+  import * as router from 'svelte-spa-router';
+
+  export let title: string | undefined = undefined;
```
brandonbloom, line 5: nit: `= undefined` can just be omitted:

```typescript
  export let title: string | undefined;
```

brandonbloom (approved):
LGTM, though I wonder why not enhance NavbarButton rather than wrapping it?

gui/src/components/nav/NavbarRoute.svelte
jwmza: Good to know, thanks :)

jwmza:
> I wonder why not enhance NavbarButton rather than wrapping it?

Could do it either way but I figured this was more elegant to allow a combination of link and custom click behaviors. Happy to change it to the other approach if there's a reason to.

PR #306, opened 2021-09-08
Create container asynchronously

kendru:
- When a container is created, do not wait for initialization to complete. Instead, initialize inside a job, and return the job ID.
- When pulling a container, do not print the status to `stdout`. Instead, report task progress and status messages.
- Fix marshalling of `ServiceNetwork` so that it can be re-parsed from the spec.


cmd/exo/new_container.go
```diff
@@ -166,7 +166,6 @@ exo flags and options for your docker container command.
 		if err != nil {
 			return err
 		}
-		fmt.Println(output.ID)
-		return nil
+		return watchJob(ctx, cl.Kernel(), output.JobID)
```
brandonbloom, line 169: there is also `new_process.go` and `new_volume.go`, etc that should be updated.

internal/core/server/workspace.go
```diff
@@ -433,16 +433,27 @@ func (ws *Workspace) getEnvironment(ctx context.Context) (map[string]string, err
...
-		Spec: input.Spec,
-	}, id)
-	if err != nil {
-		return nil, err
-	}
+	job := ws.TaskTracker.StartTask(ctx, "creating "+input.Name)
+
+	go func() {
+		defer job.Finish()
+		err := ws.createComponent(ctx, manifest.Component{
```
brandonbloom, line 440: Hm, I would not have thought to do it this way - it does work, but it introduces a new problem. I was expecting that `createComponent` would get split in to synchronous and asynchronous parts, so that errors could be returned synchronously for spec changes where those failures can be detected up front. Primarily: name conflicts.

brandonbloom (approved):
LGTM as a step forward, but some follow-up work for us to do.

internal/core/server/workspace.go
kendru: I forgot about name conflicts. My assumption was that in practice creating the spec would not fail, and this would be "good enough" until the component model gets reworked, but I don't think that's the case. I'll go ahead and make this change.

cmd/exo/new_container.go
kendru: Done

PR #307, opened 2021-09-08
Document the release/upgrade process

kendru:
This PR only adds documentation explaining how the release and upgrade process works. It does not contain any functional changes.


brandonbloom (approved):
Thanks!

PR #315, opened 2021-09-09
Use actual links in Compose create form help

jwmza:
Takes the links out of the codeblocks and puts them into actual links in the show/hide help sections of the Container, Volume, and Network creation forms.

image

### DIFF
--- a/gui/src/pages/NewDockerNetwork.svelte
+++ b/gui/src/pages/NewDockerNetwork.svelte
@@ -11,12 +11,10 @@
 </script>
 
 <NewDockerComponent {params} componentType="network" displayType="Network">
-  <div slot="help">
-    <p>
-      See <a {href}>compose network reference</a>.
-    </p>
-    <CodeBlock>
-      {specExample}
-    </CodeBlock>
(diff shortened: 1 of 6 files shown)


jwmza:
image

gui/src/pages/NewDockerNetwork.svelte
```diff
@@ -1,16 +1,16 @@
...
+  import CodeBlock from '../components/CodeBlock.svelte';
   import NewDockerComponent from './NewDockerComponent.svelte';
   import type { Params } from './NewDockerComponent.svelte';
 
   export let params: Params;
 
-  const specExample = `# See https://github.com/compose-spec/compose-spec/blob/master/spec.md#Networks-top-level-element
-name: mynetwork`;
+  const code = `name: mynetwork`;
+  const href =
```
brandonbloom, line 9: nit: a better name for this would be something like `docsHref` or `docsUrl` or like that, since absent any qualifiers, it isn't clear what the href would mean.

brandonbloom (approved):

PR #316, opened 2021-09-09
fix shell form docker commands with image pull policy

brandonbloom:


internal/providers/docker/components/container/image.go
```diff
@@ -15,29 +15,35 @@ func (c *Container) ensureImage(ctx context.Context) error {
...
-	}
-	if inspection.ID == "" {
-		if err := c.pullImage(ctx); err != nil {
-			return fmt.Errorf("pulling image: %w", err)
+	if c.canBuild() {
+		if err := c.buildImage(ctx); err != nil {
+			return fmt.Errorf("building image: %w", err)
 		}
-		inspection, _, err = c.Docker.ImageInspectWithRaw(ctx, c.Spec.Image)
+		inspection, _, err = c.Docker.ImageInspectWithRaw(ctx, c.State.Image.ID)
```
brandonbloom, line 24: This is a tad messy. I'm starting to think/work on the component model a bunch. I think I'm going to treat images as a sub-component of containers in the future, should simplify the build/use/state logic a bit.

BenElgar (commented):
LGTM

PR #319, opened 2021-09-09
handle docker.NotFound in dispose lifecycle methods

brandonbloom:
Containers already did this, now volumes and networks behave the same way.

### DIFF
--- a/internal/providers/docker/components/network/lifecycle.go
+++ b/internal/providers/docker/components/network/lifecycle.go
@@ -8,6 +8,7 @@ import (
 	core "github.com/deref/exo/internal/core/api"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/filters"
+	docker "github.com/docker/docker/client"
 )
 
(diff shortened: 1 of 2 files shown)


internal/providers/docker/components/network/lifecycle.go
```diff
@@ -82,7 +83,12 @@ func (n *Network) Dispose(ctx context.Context, input *core.DisposeInput) (*core.
 	if n.NetworkID == "" {
 		return &core.DisposeOutput{}, nil
 	}
-	if err := n.Docker.NetworkRemove(ctx, n.NetworkID); err != nil {
+	err := n.Docker.NetworkRemove(ctx, n.NetworkID)
+	if docker.IsErrNotFound(err) {
+		n.Logger.Infof("disposing network not found: %q", n.NetworkID)
```
BenElgar, line 88: Nit but I ran into this before I saw this PR and didn't understand what a "disposing volume" was. Would prefer something like `volume to be removed cannot be found`.

brandonbloom: ok fixed in bb52c7e

PR #320, opened 2021-09-09
stable dexo binary

brandonbloom:
The previous approach of using a trap to cleanup temporary files is
broken with respect to running supervise or other re-entrant commands
that run in the daemon and rely on os.Args[0] existing after the initial
command returns.

Explicitly using -o and the .dev directory won't leak temp files, which
is why the trap was previously included.


BenElgar (approved):
LGTM

Issue #270, opened 2021-09-01
[BUG] Cannot recover from bad spec


BenElgar:
**This is only an issue on the interpolation branch.**

If a user applies a docker compose spec that is invalid in such a way that it cannot be unmarshalled this cannot be recovered from. This is because the spec is saved in the state file before it is validated and every future attempt to manipulate the spec attempts to unmarshal this bad spec, which fails.

To recreate, apply a spec like this:
```docker-compose.yaml
services:
  t0:
    image: bash
    command: sleep infinity
    ports:
      - "443:" # This is an invalid port specification
```

Running this with `dexo run` will appropriately fail with an `invalid port mapping syntax` error. The problem is that even if the invalid specification is fixed, future commands will fail until the spec is manually removed from the state file:
```
deleting frontend unmarshalling spec: invalid port mapping syntax
```

The solution is presumably to validate the spec before writing.

brandonbloom:
Yikes. Good catch.

Somewhat related, I've opened #271

> This is because the spec is saved in the state file before it is validated and every future attempt to manipulate the spec attempts to unmarshal this bad spec, which fails.

Subtle correction: I think it is failing on _all_ attempts to invoke controller methods, not just future/subsequent attempts after the first one.

> The solution is presumably to validate the spec before writing.

This might be part of the solution, but I don't think it is a complete solution. In theory, even if the spec is value now, it can become invalid later. For example, if the environment variables change. Rather than try to avoid writing bad components, I'd rather be more robust to errors in bad components. For example, if a component fails to initialize, then we should ignore the invalid spec on a component update, skipping any dispose step, and treating it like a fresh initialize.

This brings up a related issue, that the procfile stuff already suffers from: Variable interpolation means that if we don't store the variable state, we actually can't accurately reproduce the previous spec on an update. Ideally, we'd automatically capture the variable bindings that were used during interpolation, so that subsequent updates can correctly diff.

brandonbloom:
> This is only an issue on the interpolation branch.

Is that actually true? or does that branch only make it worse?

I'd really like to land #266, but this seems like the blocker to fix first.

BenElgar:
Yes, I believe so. On master the given example fails at `loadManifest` but on the interpolation branch `loadManifest` succeeds and, as a result, the bad spec is written to the state file, rendering it broken.

https://github.com/deref/exo/blob/98527c58795bfe2cb96515d190715982b804564d/internal/core/server/workspace.go#L105

[...]

PR #321, opened 2021-09-10
`exo edit` command to edit component specs

brandonbloom:
In addition to being generally useful, this is a bandaid solution to #270 good enough to allow us to land #266.


BenElgar (approved):
LGTM but I don't know that I buy that this is sufficient to land #266. I ran into the bug within seconds of trying to use interpolation and did so multiple times. I'd be surprised if anyone could make non-trivial use of the feature without running into the bug. If we documented the issue and the solution then it might be okay but right now all of these are hidden commands.

brandonbloom:
Ah, I did not intend for this one to be hidden! Bad copy/paste, will change that.

Fair point RE: 266, I'll keep working at it.

PR #322, opened 2021-09-10
Simplify supervisor logging

BenElgar:
This simplifies the supervisor logging slightly by using `ReadString` instead of `ReadLine` so that the prefixes are handled automatically. It also ensures that the log message is always written, even if an error occurs.


brandonbloom (commented):
Usage of ReadLine in preference to ReadString was intentional, since ReadString will perform unbounded buffering. When I coded this up the first time, I tested with a program that just printed in a loop without writing any newline characters. The result was predictable: the supervisor process consumes all of memory!

Sounds like you this PR is trying to solve two problems:

1) truncated messages
2) some error case?

for (1), we should probably forcibly "wrap", that is, write multiple messages, inserting artificial line breaks. That's what most line-oriented log systems do in my experience.

could you elaborate on (2) please?

BenElgar:
Ah fair enough. Though honestly I think unbounded buffering is better than unexpected truncation. I have definitely had programs correctly emit several megabytes of JSON as log lines (e.g. a max size MongoDB object) and I would want that to continue to work and not be truncated or split. I have never had a log line be so long that it got close to consuming a noticeable percentage of memory. But I think either case is rare enough that I'm happy to revert that change if I haven't persuaded you.

This PR started as me trying to fix the duplicate line issue before I realised it was only happening for docker. Right now the only functional change is always emitting log messages — though this is important as before we were throwing away the last line if not terminated with a new line character.

brandonbloom:
Most syslog implementations have relatively small line limits. Some as small as 1024, frequently one memory page size such as 4096, and sometimes a megabyte or so. Every major logging/tracing system simply truncates messages that are too long. For instance, Splunk defaults to 10,000 characters. Honeycomb limits whole event sizes to 100KB and string fields to 64KB.

We're not trying to handle arbitrary stdout/stderr, we're trying to handle log messages. I think truncation is a perfectly reasonable thing to do, though we should probably capture the fact that truncation occurred in the event somehow, so we can indicate that in the output, so folks don't wonder "where did the end of this message go?"

The issue isn't a log line that is so long it consumes a noticeable percentage of memory, the issue is _bugs_ where somebody forgets a "\n" in a print statement, and then the program just runs forever building up one big line.

> though this is important as before we were throwing away the last line if not terminated with a new line character.

Can we fix this by treating eof/closed errors as prefix=true?

BenElgar:
Rejigged it to restore most of the original logic but to handle errors after processing any potential message — assumption being that if the message is non-zero we want to record it.

internal/supervise/main.go
```diff
@@ -156,41 +156,55 @@ func Main() {
...
-		if errors.Is(err, io.EOF) || errors.Is(err, os.ErrClosed) {
-			return
-		}
-		if err != nil {
-			fatalf("reading %s: %v", name, err)
-		}
-		// TODO: Do something better with lines that are too long.
 		for isPrefix {
 			// Skip remainder of line.
 			message = append([]byte{}, message...)
```
brandonbloom, line 166: blame suggests I wrote this line, but I have no idea what it is trying to do... Why is it copying the message, but only if `isPrefix` seems like this line could just be deleted?

brandonbloom (approved):

PR #323, opened 2021-09-10
Strip new lines from docker

BenElgar:
This strips trailing new line characters received as logs from the docker daemon, fixing an issue with duplicated new lines in the UI.

Whilst this works, I still think this is the wrong fix. We're fixing a display issue by changing the way we process logs. In the same way that docker preserves the newline characters I think we should too. Now, instead of the clients having to strip the newlines, every provider will have to implement that logic.

### DIFF
--- a/internal/logd/logd.go
+++ b/internal/logd/logd.go
@@ -107,9 +107,14 @@ func syslogToEvent(syslogMessage syslog.Message) (*api.AddEventInput, error) {
 		}
 	}
 
+	message := *rfc5425Message.Message
+	if message[len(message)-1] == '\n' {
+		message = message[:len(message)-1]
+	}
+
 	return &api.AddEventInput{
 		Log:       logName,
 		Timestamp: rfc5425Message.Timestamp.Format(chrono.RFC3339MicroUTC),
-		Message:   *rfc5425Message.Message,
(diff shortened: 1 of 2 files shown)


internal/logd/logd.go
```diff
@@ -107,9 +107,14 @@ func syslogToEvent(syslogMessage syslog.Message) (*api.AddEventInput, error) {
 		}
 	}
 
+	message := *rfc5425Message.Message
+	if message[len(message)-1] == '\n' {
```
brandonbloom, line 111: `message = strings.TrimSuffix(message, "\n")`

brandonbloom (approved):

internal/logd/logd.go
brandonbloom: especially because if this is possibly false, then it implies that message == "" is possibly true, which means this will be an index-out-of-bounds panic

brandonbloom:
iirc there was also a hack done in the GUI - can we undo that as part of this PR?

internal/logd/logd.go
BenElgar: Ah! That's what I was looking for! I only found `TrimRight`. Thanks!

PR #328, opened 2021-09-13
Fix error with async image pull

kendru:
This PR fixes the issue reported by @BenElgar in Slack where images that were not already present on the host caused the CreateContainer task to fail. This happened because the task was handed a context that would be cancelled by the time we used it to try to download a new image.

### DIFF
--- a/cmd/exo/root.go
+++ b/cmd/exo/root.go
@@ -26,9 +26,7 @@ For more information, see https://exo.deref.io`,
 }
 
 // newContext creates a global context that is used as the top-level process context.
-// All request-specific contexts are derived from it. Any key added to the context
-// here needs to have a corresponding call to contextutil.RegisterContextCloner so
-// that we can copy these values from a request-specific context to an asynchronous one.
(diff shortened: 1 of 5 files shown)


BenElgar (approved):
Brilliant, thank you! LGTM — but I know little about about the best practices surrounding Go contexts.

brandonbloom:
Interesting... I've seen the following instead (via https://stackoverflow.com/a/54132324):

```go
import (
    "context"
    "time"
)

type noCancel struct {
    ctx context.Context
}

func (c noCancel) Deadline() (time.Time, bool)       { return time.Time{}, false }
func (c noCancel) Done() <-chan struct{}             { return nil }
func (c noCancel) Err() error                        { return nil }
func (c noCancel) Value(key interface{}) interface{} { return c.ctx.Value(key) }

// WithoutCancel returns a context that is never canceled.
func WithoutCancel(ctx context.Context) context.Context {
    return noCancel{ctx: ctx}
}
```

PR #331, opened 2021-09-14
Amplitude analytics

kendru:
Hook up telemetry to send to Amplitude Analytics:
- Maintain a device ID that is used as a unique identifier for all events from a particular installation.
  - Send this device ID in the upgrade requests so that the API that serves the install script can attribute the upgrade to that user.
- Report very basic system information on startup (os, architecture, cpu count)
- Update the default telemetry implementation to use Amplitude instead of the previous bespoke solution.

### DIFF
--- a/AMPLITUDE_API_KEY
+++ b/AMPLITUDE_API_KEY
@@ -1,0 +1,1 @@
+Since the API key is semi-private, this file is replaced in CI with the real API key.
--- a/cmd/exo/root.go
+++ b/cmd/exo/root.go
@@ -31,7 +31,12 @@ func newContext() context.Context {
 	logger := logging.Default()
 	ctx = logging.ContextWithLogger(ctx, logger)
 
-	tel := telemetry.New(ctx, &cfg.Telemetry)
+	// This telemetry object will be replaced by one that includes the
+	// installation ID when the daemon starts. For one-off commands,
+	// it is fine for the ID to not be populated.
+	tel := telemetry.New(ctx, telemetry.Config{
+		Disable: cfg.Telemetry.Disable,
+	})
 	ctx = telemetry.ContextWithTelemetry(ctx, tel)
 
 	return ctx
--- a/cmd/exo/upgrade.go
+++ b/cmd/exo/upgrade.go
@@ -2,11 +2,15 @@ package main
 
 import (
 	"fmt"
+	"path/filepath"
 	"strings"
 
 	"github.com/deref/exo"
+	state "github.com/deref/exo/internal/core/state/api"
+	"github.com/deref/exo/internal/core/state/statefile"
 	"github.com/deref/exo/internal/telemetry"
 	"github.com/deref/exo/internal/upgrade"
+	"github.com/deref/exo/internal/util/cmdutil"
 	"github.com/spf13/cobra"
 )
 
@@ -46,7 +50,17 @@ var upgradeCmd = &cobra.Command{
 		case -1:
 			fmt.Println("Upgrade needed")
 			// TODO: Prompt for confirmation?
-			return upgrade.UpgradeSelf()
+			statePath := filepath.Join(cfg.VarDir, "state.json")
+			deviceIDPath := filepath.Join(cfg.VarDir, "deviceid")
+			store := statefile.New(statefile.Config{
+				StoreFilename:    statePath,
+				DeviceIDFilename: deviceIDPath,
+			})
+			ensureDeviceOut, err := store.EnsureDevice(ctx, &state.EnsureDeviceInput{})
+			if err != nil {
+				cmdutil.Fatalf("getting device version")
+			}
+			return upgrade.UpgradeSelf(ensureDeviceOut.DeviceID)
 		case 1:
 			fmt.Println("You are already running a prerelease version; not downgrading.")
 		default:
--- a/internal/core/state/api/store.josh.hcl
+++ b/internal/core/state/api/store.josh.hcl
@@ -64,6 +64,8 @@ interface "store" {
 
   method "ensure-installation" {
     doc = "Ensure that one-time initialization has been completed. This is safe to call whenever the server is restarted."
+
+    output "installation-id" "string" {}
   }
 }
 
--- a/internal/core/state/api/store.josh.hcl
+++ b/internal/core/state/api/store.josh.hcl
@@ -62,6 +62,9 @@ interface "store" {
     input "id" "string" {}
   }
 
+  method "ensure-installation" {
+    doc = "Ensure that one-time initialization has been completed. This is safe to call whenever the server is restarted."
+  }
 }
 
 struct "workspace-description" {
--- a/internal/core/state/api/store.josh.hcl
+++ b/internal/core/state/api/store.josh.hcl
@@ -62,10 +62,10 @@ interface "store" {
     input "id" "string" {}
   }
 
-  method "ensure-installation" {
-    doc = "Ensure that one-time initialization has been completed. This is safe to call whenever the server is restarted."
+  method "ensure-device" {
+    doc = "Ensure that one-time device initialization has been completed. This is safe to call whenever the server is restarted."
 
-    output "installation-id" "string" {}
+    output "device-id" "string" {}
   }
 }
 
--- a/internal/core/state/statefile/statefile.go
+++ b/internal/core/state/statefile/statefile.go
@@ -12,6 +12,7 @@ import (
 	"github.com/deref/exo/internal/core/state/api"
 	state "github.com/deref/exo/internal/core/state/api"
 	"github.com/deref/exo/internal/deps"
+	"github.com/deref/exo/internal/gensym"
 	"github.com/deref/exo/internal/util/atom"
 	"github.com/deref/exo/internal/util/errutil"
 	"github.com/deref/exo/internal/util/pathutil"
@@ -32,6 +33,7 @@ var _ state.Store = (*Store)(nil)
 type Root struct {
 	Workspaces          map[string]*Workspace `json:"workspaces"`          // Keyed by ID.
 	ComponentWorkspaces map[string]string     `json:"componentWorkspaces"` // Component ID -> Workspace ID.
+	InstallationID      string                `json:"installationId"`
 }
 
 type Component struct {
@@ -442,4 +444,18 @@ func (sto *Store) RemoveComponent(ctx context.Context, input *state.RemoveCompon
 		return nil, err
 	}
 	return &state.RemoveComponentOutput{}, nil
+}
+
+func (sto *Store) EnsureInstallation(ctx context.Context, input *state.EnsureInstallationInput) (*state.EnsureInstallationOutput, error) {
+	_, err := sto.swap(func(root *Root) error {
+		if root.InstallationID != "" {
+			return nil
+		}
+		root.InstallationID = gensym.RandomBase32()
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	return &state.EnsureInstallationOutput{}, nil
 }
--- a/internal/core/state/statefile/statefile.go
+++ b/internal/core/state/statefile/statefile.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"os"
 	"path/filepath"
 	"sort"
 	"strings"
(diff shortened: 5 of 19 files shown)


internal/upgrade/autoinstalled.go
```diff
@@ -15,14 +15,14 @@ import (
...
-func UpgradeSelf() error {
+func UpgradeSelf(deviceID string) error {
 	tmpfile, err := ioutil.TempFile("", "example")
 	if err != nil {
 		return fmt.Errorf("creating temporary file: %w", err)
 	}
 	defer os.Remove(tmpfile.Name())
 
-	resp, err := http.Get(exo.UpdateScriptEndpoint)
+	resp, err := http.Get(fmt.Sprintf("%s?id=%s&prev=%s", exo.UpdateScriptEndpoint, deviceID, exo.Version))
```
brandonbloom, line 25: security: since the device id comes from a file, we should do a url-encode here

AMPLITUDE_API_KEY
```diff
@@ -0,0 +1 @@
+Since the API key is semi-private, this file is replaced in CI with the real API key.
```
brandonbloom, line 1: the API key is in plain text on every deployed instance right? seems like the reason to replace it in CI is less about it being semi-private than it is about it being configurable ie if somebody used a free-and-open-only distro of exo. Is my understanding correct?

cmd/exo/root.go
```diff
@@ -31,7 +31,12 @@ func newContext() context.Context {
 	logger := logging.Default()
 	ctx = logging.ContextWithLogger(ctx, logger)
 
-	tel := telemetry.New(ctx, &cfg.Telemetry)
+	// This telemetry object will be replaced by one that includes the
+	// installation ID when the daemon starts. For one-off commands,
```
brandonbloom, line 35: installation id or device id? two different things?

cmd/exo/upgrade.go
```diff
@@ -46,7 +50,17 @@ var upgradeCmd = &cobra.Command{
...
-			return upgrade.UpgradeSelf()
+			statePath := filepath.Join(cfg.VarDir, "state.json")
+			deviceIDPath := filepath.Join(cfg.VarDir, "deviceid")
+			store := statefile.New(statefile.Config{
+				StoreFilename:    statePath,
+				DeviceIDFilename: deviceIDPath,
+			})
+			ensureDeviceOut, err := store.EnsureDevice(ctx, &state.EnsureDeviceInput{})
+			if err != nil {
+				cmdutil.Fatalf("getting device version")
```
brandonbloom, line 61: did you mean "getting device id"?

internal/core/state/api/store.josh.hcl
```diff
@@ -62,6 +62,11 @@ interface "store" {
     input "id" "string" {}
   }
 
+  method "ensure-device" {
+    doc = "Ensure that one-time device initialization has been completed. This is safe to call whenever the server is restarted."
```
brandonbloom, line 66: are there times when this is not safe to call? or do you just mean it is safe to call multiple times?

internal/core/state/statefile/statefile.go
```diff
@@ -5,33 +5,43 @@ import (
...
+	"github.com/deref/exo/internal/gensym"
 	"github.com/deref/exo/internal/util/atom"
 	"github.com/deref/exo/internal/util/errutil"
 	"github.com/deref/exo/internal/util/pathutil"
 )
 
-func New(filename string) *Store {
+type Config struct {
+	StoreFilename    string
+	DeviceIDFilename string
```
brandonbloom, line 24: why store this as a separate file vs a key in the state file?

cmd/exo/upgrade.go
```diff
@@ -46,7 +50,17 @@ var upgradeCmd = &cobra.Command{
...
 			fmt.Println("Upgrade needed")
 			// TODO: Prompt for confirmation?
-			return upgrade.UpgradeSelf()
+			statePath := filepath.Join(cfg.VarDir, "state.json")
+			deviceIDPath := filepath.Join(cfg.VarDir, "deviceid")
+			store := statefile.New(statefile.Config{
+				StoreFilename:    statePath,
+				DeviceIDFilename: deviceIDPath,
+			})
+			ensureDeviceOut, err := store.EnsureDevice(ctx, &state.EnsureDeviceInput{})
```
brandonbloom, line 59: not sure i love the idea of the cli directly accessing the store api, since the intent is that the daemon manages this state

brandonbloom (commented):
I still need to look more closely at amplitude.go and telemetry/default.go, but the other stuff LGTM with one concern/question:

Should the device ID stuff go in to the "store" interface? Seems like there should just be a new object/interface like DeviceState or InstallationStore or something that only deals with that one device id file. See comments about separate files & access to store api from exo client.

WDYT?

internal/telemetry/default.go
```diff
@@ -173,14 +118,19 @@ func (t *defaultTelemetry) sendRecordedTelemetry() {
...
-		t.SendEvent(t.ctx, &OperationsPerformed{
-			Operation:       op,
-			Success:         success,
-			DurationSummary: summary,
-		})
+		t.SendEvent(t.ctx, OperationsPerformedEvent(op, success, summary))
 	}
 }
 
+func (t *defaultTelemetry) nextEventID() int {
```
brandonbloom, line 125: so event ids are scoped by session id and start at zero?

internal/telemetry/default.go
```diff
@@ -81,64 +75,15 @@ func (t *defaultTelemetry) RecordOperation(op OperationInvocation) {
...
-
-		typedRes := startSessionResponse{}
-		if err := json.NewDecoder(res.Body).Decode(&typedRes); err != nil {
-			errSetSession(res, err)
-			return
-		}
-
-		t.sessionID = typedRes.SessionID
-		logger.Infof("Started session: %q", t.sessionID)
+		t.sessionID = chrono.NowMillisecond(ctx)
```
brandonbloom, line 85: this can run in the cli right? what if somebody does `exo whatever & exo whatever else` is there a session id race?

internal/telemetry/amplitude.go
```diff
@@ -0,0 +1,128 @@
...
+	case a.events <- evt:
+		return nil
+	default:
+		return fmt.Errorf("event channel buffer of size %d is full", eventChannelBufferSize)
+	}
+}
+
+func (a *AmplitudeClient) Flush() {
+	notify := make(chan struct{})
+	defer close(notify)
```
brandonbloom, line 58: What is the purpose of this close?

AMPLITUDE_API_KEY
kendru: Yes, the API key is in plain text. Users could extract it and send bogus data to Amplitude, but it is not a secret key. The best practice is to "make it a little more difficult to misuse", but I'm fine either putting the actual data in this file or updating the placeholder text to be more descriptive. Thoughts?

internal/telemetry/amplitude.go
```diff
@@ -0,0 +1,128 @@
...
+			}
+
+		case notify := <-a.flush:
+			doPublish()
+			notify <- struct{}{}
+		}
+	}
+}
+
+func (a *AmplitudeClient) publish(events []Event) error {
```
brandonbloom, line 105: this returns an error, but nobody logs it or anything

cmd/exo/root.go
kendru: it should be device id. i did an auto refactor and missed the comments.

internal/telemetry/amplitude.go
```diff
@@ -0,0 +1,128 @@
...
+func (a *AmplitudeClient) Close() {
+	a.Flush()
+	a.cancel()
+}
+
+func (a *AmplitudeClient) run() {
+	timer := time.NewTimer(eventSendInterval)
+
+	buffer := make([]Event, eventBufferSize)
+	idx := 0
```
brandonbloom, line 74: seems like this is just len vs cap? why not `make([]Event, 0, eventBufferSize)`?

internal/core/state/statefile/statefile.go
kendru: it is generally written by the install script, so I put it in its own file to avoid JSON editing with unix text processing tools :)

cmd/exo/upgrade.go
kendru: How about if I make a kernel endpoint that returns it, then have this entrypoint call `ensureDaemon()` and make an API call?

internal/telemetry/default.go
kendru: Yes - event id is only used to deduplicate events from a single user that may have the same timestamp.

internal/telemetry/default.go
kendru: It should be fine. They will be considered one session if they execute in the same millisecond and two sessions otherwise. This is one of the quirks that I found with Amplitude.

internal/telemetry/amplitude.go
kendru: This was just copied and pasted from https://github.com/savaki/amplitude-go, but it looks like it is not necessary.

internal/telemetry/amplitude.go
kendru: More copy/paste from https://github.com/savaki/amplitude-go. I'll add a logger to this struct and at least log the errors.

cmd/exo/upgrade.go
brandonbloom: I think that the less the upgrade path relies on the daemon, the better. Maybe we ship some bug where the daemon is crashing or something & we want people to be able to upgrade to fix it!

internal/telemetry/amplitude.go
kendru: This is a buffer that is allocated once and whose head is overwritten on every flush so that we do not allocate a new buffer on every flush.

cmd/exo/upgrade.go
brandonbloom: to be clear: I think it's OK that the daemon and the CLI share code to do this, just not specifically the statefile package, since that's for managing the state of components/stacks. A package dedicated to the subset of functionality for device ids, upgrades, etc seems like a more minimal surface for bugs.

AMPLITUDE_API_KEY
brandonbloom: Is it possible to simply replace it with a dev key? This way we'd get isolated analytics for folks working on the product?

internal/core/state/statefile/statefile.go
brandonbloom: gotcha. that makes good sense :)

internal/telemetry/default.go
brandonbloom: ok, so sessions are not durable or persistent across multiple invocations of the CLI for example?

internal/telemetry/amplitude.go
brandonbloom: right, i'm assuming this is just more "copy/paste dfrom amplitude-go" things - in this case, i think it `idx` in `buffer` is serving the same role as `len(buffer)` where the `cap(buffer) == eventBufferSize`

internal/core/state/statefile/statefile.go
```diff
@@ -32,6 +36,7 @@ var _ state.Store = (*Store)(nil)
 type Root struct {
 	Workspaces          map[string]*Workspace `json:"workspaces"`          // Keyed by ID.
 	ComponentWorkspaces map[string]string     `json:"componentWorkspaces"` // Component ID -> Workspace ID.
+	DeviceID            string                `json:"deviceId"`
```
brandonbloom, line 39: leftovers?

internal/core/state/statefile/statefile.go
```diff
@@ -17,9 +17,13 @@ import (
 	"github.com/deref/exo/internal/util/pathutil"
 )
 
-func New(filename string) *Store {
+type Config struct {
+	StoreFilename string
+}
+
+func New(cfg Config) *Store {
```
brandonbloom, line 24: can probably revert this `Config` change

internal/exod/main.go
```diff
@@ -93,7 +91,24 @@ func RunServer(ctx context.Context, flags map[string]string) {
 	}
 
 	statePath := filepath.Join(cfg.VarDir, "state.json")
-	store := statefile.New(statePath)
+	// deviceIDPath := filepath.Join(cfg.VarDir, "deviceid")
```
brandonbloom, line 94: leftovers

internal/util/atom/file.go
```diff
@@ -32,6 +32,7 @@ func (a *FileAtom) Deref(v interface{}) error {
 
 	bs, err := ioutil.ReadFile(a.filename)
 	if os.IsNotExist(err) {
+		// XXX [ATOM JSON CODING]: This assumes a JSON codec.
```
brandonbloom, line 35: whoops!

brandonbloom (commented):
Thanks @kendru. As discussed, I'll take it from here.

internal/core/state/statefile/statefile.go
kendru: Yep...

internal/util/atom/file.go
kendru: :D

PR #333, opened 2021-09-14
Handle empty syslog messages.

### LINKED ISSUE #326: [BUG] Empty log lines not captured
**Describe the bug**
Logs that consist of nothing other than a new line character are not captured.

**To Reproduce**
Run either the docker compose or Procfile below:
```yaml
services:
  t0:
    image: bash
    command: sh -c 'i=0; while true; do echo; sleep 1; done'
```
```
t0: sh -c 'i=0; while true; do echo; sleep 1; done'
```

Neither will emit any logs and the Procfile will cause the following error in the server:
```
13:14:58    server 2021/09/13 13:14:58 interpreting syslog message: expected MSG
```

**Expected behavior**
Exo should log lots of empty log lines.

**Additional context**
My understanding is that it is a violation of the procotol to send empty messages so, by removing new line characters, we are creating invalid syslog packets.
> There is no minimum length of the syslog message although sending a syslog packet with no contents is worthless and SHOULD NOT be transmitted.

https://datatracker.ietf.org/doc/html/rfc3164#section-4.1

brandonbloom:
Fixes #326

### DIFF
--- a/internal/logd/logd.go
+++ b/internal/logd/logd.go
@@ -81,9 +81,6 @@ func syslogToEvent(syslogMessage syslog.Message) (*api.AddEventInput, error) {
 	if rfc5425Message.Appname == nil {
 		return nil, errors.New("expected APP-NAME")
 	}
-	if rfc5425Message.Message == nil {
-		return nil, errors.New("expected MSG")
-	}
 	if rfc5425Message.MsgID == nil {
 		return nil, errors.New("expected MSGID")
 	}
@@ -108,7 +105,11 @@ func syslogToEvent(syslogMessage syslog.Message) (*api.AddEventInput, error) {
 		}
 	}
 
-	message := strings.TrimSuffix(*rfc5425Message.Message, "\n")
+	message := ""
+	if rfc5425Message.Message != nil {
+		message = strings.TrimSuffix(*rfc5425Message.Message, "\n")
+	}
+
 	return &api.AddEventInput{
 		Log:       logName,
 		Timestamp: rfc5425Message.Timestamp.Format(chrono.RFC3339MicroUTC),


BenElgar (approved):
I still think that empty messages are invalid according to the spec — but I guess it doesn't matter if we violate it given we're both sender and receiver :man_shrugging:

brandonbloom:
Why do you still think that? I'm looking at the spec and it says:

> MSG Part of a syslog Packet
>
>   The MSG part will fill the remainder of the syslog packet.  This will
>   usually contain some additional information of the process that
>   generated the message, and then the text of the message.  **There is no
>   ending delimiter to this part.**

Emphasis mine.

Additionally:

> The MSG part has two fields known as the TAG field and the CONTENT
>   field.  The value in the TAG field will be the name of the program or
>   process that generated the message.  The CONTENT contains the details
>   of the message.

In the examples below, it says:

> each message has been indented, **with line breaks inserted in this document for readability.

To me, this all means that (1) empty _CONTENT_ is perfectly valid. It's just empty packets that are not valid and (2) the content is not expected to have a newline terminator.

BenElgar:
Sorry, I wrote that before having seen your comment on the issue.

TLDR: I'll shut up :slightly_smiling_face: 

So the spec we were using, RFC3164, we were using says
> The MSG part of the syslog packet MUST contain visible (printing) characters

which seemed pretty clear to me. I suppose you could argue that the spec is wrong and should read "all characters in the MSG part of the syslog packet MUST be visible (printing) characters" — but that's not what it says.

BUT, on closer inspection, that spec has been obsoleted by RFC5424, which is much clearer (and what we're actually using) and contains the following BNF:
```
      MSG             = MSG-ANY / MSG-UTF8
      MSG-ANY         = *OCTET ; not starting with BOM
```
The `*` in `*OCTET` indicates zero or more repetitions — so an empty message is indeed valid.

https://datatracker.ietf.org/doc/html/rfc3164#section-4.1.2
https://datatracker.ietf.org/doc/html/rfc5424#section-6

PR #334, opened 2021-09-15
Clean up repetitive GUI code

jwmza:
This is just a quick cleanup pass to reduce repetitive GUI code (mostly in form pages) prior to adding more GUI CRUD functionality

### DIFF
--- a/gui/src/components/form/SubmitButton.svelte
+++ b/gui/src/components/form/SubmitButton.svelte
@@ -1,0 +1,16 @@
+<script lang="ts">
+  import Button from '../Button.svelte';
+</script>
+
+<div class="buttons">
+  <Button type="submit"><slot /></Button>
+</div>
+
+<style>
+  .buttons {
+    display: flex;
+    flex-direction: row;
+    justify-content: flex-end;
+    margin: 24px 0;
+  }
+</style>
--- a/gui/src/components/form/SubmitButton.svelte
+++ b/gui/src/components/form/SubmitButton.svelte
@@ -1,15 +1,16 @@
 <script lang="ts">
   import Button from '../Button.svelte';
+
+  // This is a right-aligned submit button for forms.
+  // Used for create/edit forms where the code below
+  // would otherwise be highly repetitive.
 </script>
 
-<div class="buttons">
-  <Button type="submit"><slot /></Button>
-</div>
+<div><Button type="submit"><slot /></Button></div>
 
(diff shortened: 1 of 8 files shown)


gui/src/components/form/SubmitButton.svelte
```diff
@@ -0,0 +1,17 @@
+<script lang="ts">
+  import Button from '../Button.svelte';
+
+  // This is a right-aligned submit button for forms.
```
brandonbloom, line 4: aside: is the appropriate place for file comments? Seems like maybe they belong in a  block above the script tag? or as // comments before the imports? no sure what the Svelte conventions are.

gui/src/components/form/SubmitButton.svelte
```diff
@@ -0,0 +1,17 @@
...
+  // Used for create/edit forms where the code below
+  // would otherwise be highly repetitive.
+</script>
+
+<div><Button type="submit"><slot /></Button></div>
+
+<style>
+  div {
+    display: flex;
+    justify-content: flex-end;
```
brandonbloom, line 14: Why is this part of the button? Seems like the layout logic should be part of the form/container, such as a slot for form buttons. Consider what happens if you want _two_ buttons. How would that work? Generally, it's better to have a little bit more verbosity than it is to bake policy in too deeply into the component hierarchy.

jwmza: As it is now, this is supposed to be like a "block" component for forms that can be used and replaced with ease.

The original approach, when this `div` was called `.buttons`, was to support multiple buttons but I don't think it's worthwhile setting that up until we actually have a need for it.

It's likely that I'll refactor this in the next week with **UPDATE** forms for components, where we probably want both a primary "save changes" and secondary "cancel" and/or even "delete" buttons.

Anyways, this PR is intended to clean up the current state of things and tie up loose ends before adding full CRUD GUI which is one of the handful of priorities I have right now.

gui/src/components/form/SubmitButton.svelte
jwmza: Not sure, there's something off-putting about the `<!-- HTML style comments -->` and the way they are syntax highlighted plus it seems strange to not have imports at the top for me.

Happy to adopt whatever standard you'd like to establish here.

brandonbloom (approved):

PR #335, opened 2021-09-17
Improve `ProcessList` / Component List

jwmza:
## Before

This is better now, but still has some problems.

### The problems with this implementation

- We should not have to use `describeWorkspaces()` to get the root directory of the active workspace.
  - My main issue here is we are sending the entire list to the client and selecting the correct one there. This seems like a bad idea, if you assume the number of workspaces and detail of each workspace description grows - instead it would be useful to either have a singular `describeWorkspace(workspaceId)` which takes one `id` and returns one workspace description, or an AWS API-esque optional parameter to our read functions like `describeWorkspaces({ workspaceId })` where the normal `describeWorkspaces()` still returns the full list, but giving an `id` returns a list of length 1.
  - Update: I _think_ @BenElgar has fixed this issue in #369 . Presumably we can use `workspace.describeSelf()` instead.
- Arguably, due to the nature of the local exo application it would make sense to have such important information readily available from any page in the app without using promises. Not sure what approach would make the most sense here. 
    -  A strategy here could be to use `localStorage` to store a cache of e.g. workspace and component descriptions, so that the frontend can snap between pages super fast without visual loading and re-hydrate with real server data in the background. The difference here would be small in the local exo application (I'm assuming `localStorage` API calls are very fast, but the local exo server API calls should also be fairly fast. However, the big difference here comes if you're using exo remotely through a web app, where the server would be possibly hundreds of milliseconds delayed and `localStorage` probably <1 ms.
- The `Panel` could perhaps expose a better way to override the whole header part in cases like this where we want to control it in a way that is not just a usual text title + optional icon button.
- _~~We should not have to `{#await ...}` the root directory of the active workspace, especially when you have navigated to this page by clicking a button which already had it displayed.~~ Update: changed my mind_
- _~~The way we format and display workspace names should probably use a smart approach that finds the shortest unique path-tail among all your exo workspaces, or just the name of the root directory itself, depending on context.~~ Update: fixed, thanks @brandonbloom_


brandonbloom:
Maybe we should add a `displayName` property to the results of `describe-workspaces` & move this shortest path logic server side.

@jwmza Would that do the trick?

jwmza:
@brandonbloom Yeah I think that solves the problems

brandonbloom:
OK, you should now have the backend support you need to make this work. LMK when you're ready for another review :)

jwmza:
Will do, thanks @brandonbloom

jwmza:
I would merge this w/o review now and continue on to the cards, but I'll just do another branch on top of this instead and we can collapse them all hopefully tomorrow.

jwmza:
I've started adding in the CRUD cards + dropdown menu stuff here so it's not ready for review again.

jwmza:
Dropdown menus working well:

image

jwmza:
Still IPR, the core functionality of this screen is ~100% done, but the code is a mess and some components and styling should be generalized before merging.

gui/src/components/processes/ProcessListTable.svelte
```diff
@@ -45,131 +51,197 @@
...
-            workspaceId,
-          )}/processes/${encodeURIComponent(id)}`}>{name}</a
-        ></td
+{#each data as { id, name, running } (id)}
+  <div class="card" style={logStyleFromHash(name + ':out')}>
+    <div>
+      <ProcessRunControls {setProcRun} {statusPending} {id} {running} />
+    </div>
+
+    <div>
```
brandonbloom, line 60: Would be good to extract these cards into their own component files. Smaller components are usually better.

brandonbloom (approved):
LGTM

brandonbloom:
I know you want to generalize styles etc. I think all this code is safe to go to main as is - but definitely do follow-up work to improve the code.

gui/src/components/processes/ProcessListTable.svelte
jwmza: Yeah, this is part of the mentioned cleanup :) thanks for the review

PR #337, opened 2021-09-20
New Project UI

jwmza:
This PR will contain GUI pages to enable the creation of new Projects entirely within the GUI, including:

- Selection of a "starter" or "template" e.g. "Empty project" or "Next+Prisma+Postgres app"
- Selection of a parent directory + naming of the project (also the new directory name)
- Triggering the underlying creation process and showing a loading screen until the new project is ready

### DIFF
--- a/gui/src/routes.ts
+++ b/gui/src/routes.ts
@@ -1,5 +1,7 @@
 import Home from './pages/Home.svelte';
-import NewWorkspace from './pages/NewWorkspace.svelte';
+import NewWorkspace from './pages/NewWorkspace.svelte'; // Delete this after New Project flow is ready
+import NewProject from './pages/NewProject.svelte';
+import NewProjectConfig from './pages/NewProjectConfig.svelte';
 import Workspace from './pages/Workspace.svelte';
 import WorkspaceComponents from './pages/WorkspaceComponents.svelte';
 import WorkspaceVariables from './pages/WorkspaceVariables.svelte';
@@ -18,7 +20,9 @@ import NotFound from './pages/NotFound.svelte';
 
 export default {
   '/': Home,
-  '/new-workspace': NewWorkspace,
+  '/new-workspace': NewWorkspace, // Delete this after New Project flow is ready
+  '/new-project': NewProject,
+  '/new-project/:starter': NewProjectConfig,
   '/workspaces/:workspace': Workspace,
   '/workspaces/:workspace/components': WorkspaceComponents,
   '/workspaces/:workspace/variables': WorkspaceVariables,


jwmza:
This now actually works to browse and select a directory from your filesystem.

The only remaining blocker is the completion of #343 , at which point I'll swap out the fake API I built to list templates here and also add in `createProject` submission/routing/error handling.

image

gui/src/lib/api.ts
```diff
@@ -194,6 +194,7 @@ export interface DescribeTasksInput {
 
 export interface DirectoryEntry {
   name: string;
+  path: string;
```
jwmza, line 197: @BenElgar was I right or wrong to add this here? Typescript was bugging me about it and it seems like it _is_ included in the results.

BenElgar: Yes you were, right, sorry. I added it to the back end as well: [`bf9bff7` (#343)](https://github.com/deref/exo/pull/343/commits/bf9bff7261dedcf854fd36f3ec416f6aafbc69b9)

gui/src/pages/NewProject.svelte
```diff
@@ -1,31 +1,52 @@
...
     <h1>New project</h1>
     <p>Select a starter for your new project:</p>
-    {#each starters as starter}
-      <button
-        on:click={() => {
-          router.push(`#/new-project/${starter.name}`);
-        }}>{starter.displayName}</button
-      >
-    {/each}
+    {#await api.kernel.describeTemplates()}
```
brandonbloom, line 44: I'm not actually sure how this works in Svelte. My intuition is that this will cause an API request every time the component is rendered, but maybe it only causes an API request when `api` changes?

gui/src/components/WorkspaceList.svelte
```diff
@@ -5,8 +5,10 @@
   export let workspaces: WorkspaceDescription[];
 </script>
 
+<button on:click={() => router.push('#/new-project')}> Add new project </button>
```
brandonbloom, line 8: Do we have a LinkButton or Button with href= yet?

gui/src/pages/NewProjectConfig.svelte
```diff
@@ -0,0 +1,149 @@
...
+  import ErrorLabel from '../components/ErrorLabel.svelte';
+  import SubmitButton from '../components/form/SubmitButton.svelte';
+  import CenterFormPanel from '../components/form/CenterFormPanel.svelte';
+  import { api, isClientError } from '../lib/api';
+  import * as router from 'svelte-spa-router';
+
+  export let params = { starter: '' };
+
+  const { starter } = params;
+  const isEmptyProject = starter === 'empty'; // XXX Use this with the `createProject` api to avoid a second (template URL) parameter if the project is empty.
```
brandonbloom, line 14: Can we make the second parameter required and change the type to string|null ?

gui/src/components/WorkspaceList.svelte
jwmza: Not for this styling, but I'm thinking of adding one soon, the main problem is we have similar buttons in different pages with the primary differences being padding/margin/border-radius, so those need to be parameterized.

gui/src/pages/NewProjectConfig.svelte
jwmza: I don't see why not

gui/src/pages/NewProjectConfig.svelte
```diff
@@ -0,0 +1,149 @@
...
+  let name = starter;
+  let error: Error | null = null;
+
+  let workingDirectory: string | null = null;
+
+  const setWorkingDirectory = (dir: string) => {
+    workingDirectory = dir;
+  };
+
+  (async () => {
```
brandonbloom, line 25: this is not a great pattern, since it introduces a potential race condition. Each time this code runs, it is racing with event handlers that set the working directory. Instead, consider changing the working directory to be of type Promise<string> and initializing it immediately that way.

gui/src/pages/NewProjectConfig.svelte
```diff
@@ -0,0 +1,149 @@
...
+            workingDirectory, // Currently this doesn't actually create the new directory, see XXX note above
+          );
+          router.push(`/workspaces/${encodeURIComponent(workspaceId)}`);
+        } catch (ex) {
+          if (!isClientError(ex)) {
+            throw ex;
+          }
+          error = ex;
+        }
+        // XXX Hack to address lack of GUI for applying procfiles, etc.
```
brandonbloom, line 48: @BenElgar I janked this in there for release a long while ago. I think we should make createProject do an apply automatically.

gui/src/pages/NewProject.svelte
BenElgar: Yes that's my intuition as well — though I can't immediately find a conclusive answer. My approach to this was to put the promise in the initial script tag which is run just on creation of the component. (https://svelte.dev/docs#script)

gui/src/pages/NewProjectConfig.svelte
jwmza: Yeah, it should, and also I just realized that we not only need to set up the project such that it is runnable but we might also want some intial step analogous to `npm i`

Even if the template sets up your project perfectly and adds components to exo, if you don't have dependencies installed the process will fail

jwmza: Maybe that's best left up to the user, but I'm not sure

jwmza:
Okay this PR is tested and working for both empty and templated projects. The templating is very fast, thanks @BenElgar - it feels great.

Some cleanup and DRY work left but otherwise the core functionality here is good to go.

gui/src/pages/NewProjectConfig.svelte
```diff
@@ -0,0 +1,150 @@
...
+<Layout>
+  <CenterFormPanel title={`New project: ${starter}`} backRoute="#/new-project">
+    <form
+      on:submit|preventDefault={async () => {
+        error = null;
+        let workspaceId;
+
+        try {
+          workspaceId = await api.kernel.createProject(
+            `${workingDirectory}/${name}`,
```
jwmza, line 46: The way I build the root directory in form submission here is kind of suspect. Seems to work but I know paths can be difficult - does it look alright to you @brandonbloom ?

gui/src/lib/api.ts
```diff
@@ -287,10 +287,13 @@ export const api = (() => {
...
       },
-      async createProject(root: string, templateUrl?: string): Promise<string> {
+      async createProject(
+        root: string,
+        templateUrl: string | null,
+      ): Promise<string> {
         const { id } = (await invoke('create-project', {
           root,
-          templateUrl,
+          templateUrl: templateUrl !== null ? templateUrl : undefined,
```
jwmza, line 296: This is a bit messy, are you sure it isn't cleaner to use `string | undefined` instead of `string | null`?

Alternatively we can modify the backend as well but I'm not sure how it works so I didn't want to change what was actually being submitted to the server @brandonbloom

gui/src/pages/NewProject.svelte
jwmza: I don't understand why this would be a problem in this case - can somebody please elaborate the source of concern?

BenElgar: The fear here was that on every render you'd get a new request. I added the following to the top level script tag to test it:
```
  // added this to top level script tag
  let v = 1;
  setInterval(() => v++, 1000);
```

I also added a `<p>{v}</p>` to the body.

But I only got one request. So I think it's all fine.

gui/src/pages/NewProjectConfig.svelte
BenElgar: So I actually set out to do that but I couldn't immediately see how to acquire a handle on a `Workspace` from a `Kernel`.  Will figure it out.

gui/src/pages/NewProjectConfig.svelte
jwmza: Okay, I believe this is fixed now, thanks @BenElgar

gui/src/pages/NewProjectConfig.svelte
```diff
@@ -0,0 +1,150 @@
...
+        bind:value={name}
+        name="name"
+        id="name"
+        placeholder={starter}
+        --input-width="100%"
+      />
+      <div style="height:32px" />
+      {#if workingDirectory}
+        <label for="root">Root:</label>
+        <h2><span>{workingDirectory}<span>/{name}</span></span></h2>
```
BenElgar, line 77: This breaks when you navigate up to the root directory.

gui/src/components/WorkspaceList.svelte
jwmza: I'm gonna follow up with a separate PR for that, it's not high priority. I think there's already an issue.

gui/src/pages/NewProjectConfig.svelte
BenElgar: This doesn't do the right thing for the root directory. The API returned path will never have a trailing slash except for the root directory (where the trailing slash is the entire path).

gui/src/pages/NewProjectConfig.svelte
```diff
@@ -0,0 +1,156 @@
...
+        {#await dirPromise}
+          <Spinner />
+        {:then dir}
+          {#if dir.parent !== null}
```
BenElgar, line 81: Minor, but I'd prefer disabling the button rather than removing it. Otherwise you have what happened to me where I was trying to get to the root directory by clicking the `..` button a bunch of times and I got to `/app` because when I got to the root directory the `..` button disappeared and `/app` appeared unexpectedly under my cursor.

gui/src/pages/NewProjectConfig.svelte
```diff
@@ -0,0 +1,156 @@
...
+            <button
+              on:click={() => setWorkingDirectory(String(dir.parent?.path))}
```
BenElgar, line 85: Checking whether `dir.parent` is nil would be both clearer and more correct. Here you introduce a race condition where if the user clicks on the button and then `dir` is changed and then the `on:click` handler is fired for the button press, `workingDirectory` will be set to `"undefined"`. (Obviously the chances of that happening are virtually zero — but it seems needless.)

gui/src/pages/NewProjectConfig.svelte
jwmza: Done

gui/src/pages/NewProjectConfig.svelte
jwmza: I _think_ it's fixed now, thanks @BenElgar

gui/src/pages/NewProjectConfig.svelte
jwmza: I'm not sure what you mean here @BenElgar - isn't my `{#if ...}` block doing this already?

gui/src/pages/NewProjectConfig.svelte
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        try {
+          workspaceId = await api.kernel.createProject(
+            `${withSlash(String(workingDirectory))}${name}`,
+            templateUrl,
+          );
+          router.push(`/workspaces/${encodeURIComponent(workspaceId)}`);
+        } catch (ex) {
+          if (!isClientError(ex)) {
```
BenElgar, line 59: This is a type error in Typescript 4.4. The variable `ex` will be of type `unknown` and `isClientError` takes an `Error`. You need to either cast to `any` or check if it's an error. 

https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#use-unknown-catch-variables

gui/src/pages/NewProjectConfig.svelte
jwmza: `dir.parent`, apparently, should be `string | null`, so the only reason I needed `String()` and `?.` was because VS Code doesn't seem to understand the type information that `{:then dir} {#if dir.parent !== null}` implies

gui/src/lib/api.ts
BenElgar: I can change that

gui/src/pages/NewProjectConfig.svelte
jwmza: Does this fix work? https://github.com/deref/exo/pull/337/commits/a4c80a1c508109e37cb4c72227d298cb53fb9d4b#

gui/src/pages/NewProjectConfig.svelte
jwmza: Should be fixed now

gui/src/pages/NewProject.svelte
jwmza: Okay, cool. I still don't really understand why that fear is a concern - seems like there wouldn't be many renders in the first place, nor does the function seem expensive. Concern makes sense if these assumptions are not true though.

gui/src/pages/NewProjectConfig.svelte
BenElgar: No — you're in an event handler which is a different context.

Let's say you've got a very slow computer (or a very busy one) and you're on this page. `dir.parent` is set so you can see the button. You click on the button. Now, because your computer is so slow, the event handler takes several hundred milliseconds to fire. In that time you click the button again, since it's still there since the event handler that would remove it hasn't fired yet. Now the first event handler runs, correctly navigating to the root directory and removing the button. Then the second one runs and sets your working directory to `"undefined"`, breaking the page.

gui/src/lib/api.ts
BenElgar: Actually, nevermind, you should just be able to pass `templateUrl`. `null` is a valid (preferable?) value to send to the server here.

gui/src/pages/NewProjectConfig.svelte
BenElgar: Not really — now it ignores the error if it isn't of type `Error`. What we really want to do is rethrow the error if it isn't the correct type.

Something like
```
if (!(ex instanceof Error) || !isClientError(ex)) {
throw ex;
}
error = ex;
```

gui/src/pages/NewProjectConfig.svelte
BenElgar: To be clear, when I say that I set out to do that, I mean call apply. I think the user should have the chance to do `npm i` or whatever else before we start the project for them.

gui/src/pages/NewProject.svelte
BenElgar: I mean I don't think it would ever have been a massive problem for this page as it is. But as a pattern it would be bad if every time anything changes on the page we refetch all data.

gui/src/pages/NewProjectConfig.svelte
```diff
@@ -0,0 +1,167 @@
...
+            >
+              ..
+            </button>
+          {:else}
+            <button disabled> .. </button>
+          {/if}
+          <div class="directories">
+            {#each dir.entries
+              .filter((x) => x.isDirectory)
+              .sort((x, y) => (x.name[0] > y.name[0] ? 1 : -1)) as entry}
```
BenElgar, line 97:
```suggestion
              .sort((x, y) => x.name.localeCompare(y.name)) as entry}
```

gui/src/pages/NewProjectConfig.svelte
jwmza: Okay done in 0f1e099

gui/src/pages/NewProjectConfig.svelte
jwmza: Overall I think the best template experience would be if we automated that as well.

Give me the full stack project, all hooked up together, with dependencies installed, and already running in dev mode. That way I can already play around with it and start making changes immediately.

jwmza: That being said, I can see some arguments for not doing that.

BenElgar (approved):
image

My final comment is that this handles errors badly. I've made the permission error more meaningful on master (https://github.com/deref/exo/pull/352) but you still run into the issue where if you get a permission denied error you can't leave the directory so you can't resolve the problem.

Don't think that that's important enough to prevent merging though.

jwmza:
Cool, thanks @BenElgar 

I'll improve the directory browser a bit later + add a "reset to default" button if the directory picker errors to take you back to the home directory

PR #338, opened 2021-09-20
Expose workspace display name to clients

brandonbloom:
First step towards the backend support needed for #335.


brandonbloom:
@jwmza please give this branch a try. You'll have to update the types/etc on the TypeScript/client side.

This doesn't do the actual shortest-path-suffix logic yet, but that's an easy extension I can follow-up with.

jwmza:
The code here looks good but I think I may have confused myself earlier.

If this only adds display name alongside root and id to describe workspace, how does it solve the await issue from my draft?

I think the proper solution in not awaiting the workspace display name is probably to store the workspace display name client side since it isn't in the URL, right?

brandonbloom:
If the user is allowed to name the project themselves - and presumably they will be able to at some point - or if there is any other information about a workspace/project that you're going to need, then you will have to await some data loading. Not sure that is avoidable.

jwmza:
Fair enough, I suppose I can move the await logic to only hold the minimal subset of the DOM that depends on it

jwmza (approved):

PR #339, opened 2021-09-20
shortest unique root suffix as display name

brandonbloom:
Follow-up to #338 and in support of #335

### DIFF
--- a/internal/core/state/statefile/displayname.go
+++ b/internal/core/state/statefile/displayname.go
@@ -1,0 +1,64 @@
+package statefile
+
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+)
+
+type displayNameBuilder struct {
+	tree      reversePathTree
+	separator string
+}
+
+func newDisplayNameBuilder(separator string) *displayNameBuilder {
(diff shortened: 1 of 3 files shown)


jwmza (approved):
LGTM

@brandonbloom what is the reason for having something dynamic like `filepath.Separator`? is this an OS difference across devices?

brandonbloom:
> what is the reason for having something dynamic like filepath.Separator?

I parameterized that so that I could use unix-style path separators (`/`) in the tests. Otherwise, the tests would have been broken on Windows, where the path separator is `\`.

jwmza:
> > what is the reason for having something dynamic like filepath.Separator?
> 
> I parameterized that so that I could use unix-style path separators (`/`) in the tests. Otherwise, the tests would have been broken on Windows, where the path separator is `\`.

I thought we didn't support Windows at all @brandonbloom

brandonbloom:
We likely will in the future. Mostly dropped windows support previously on a temporary basis to get an early release out.

jwmza:
Gotcha. That's good to hear.

PR #342, opened 2021-09-21
Add file browser backend

BenElgar:


internal/core/server/kernel.go
```diff
@@ -200,3 +200,32 @@ func (kern *Kernel) DescribeTasks(ctx context.Context, input *api.DescribeTasksI
...
+func (kern *Kernel) GetUserHomeDir(ctx context.Context, input *api.GetUserHomeDirInput) (*api.GetUserHomeDirOutput, error) {
+	path, err := os.UserHomeDir()
+	if err != nil {
+		return &api.GetUserHomeDirOutput{}, errors.New("could not get user home directory")
+	}
+	return &api.GetUserHomeDirOutput{Path: path}, nil
+}
+
+func (kern *Kernel) ReadDir(ctx context.Context, input *api.ReadDirInput) (*api.ReadDirOutput, error) {
+	if !path.IsAbs(input.Path) {
```
brandonbloom, line 213: As best I can tell, every usage of `path` in this file should actually be `filepath`. The `path` package is for URLs, idealized paths, etc. The `filepath` package will use the correct separator for the current OS. The two packages have largely identical interfaces otherwise.

internal/core/server/kernel.go
```diff
@@ -200,3 +200,32 @@ func (kern *Kernel) DescribeTasks(ctx context.Context, input *api.DescribeTasksI
 	}
 	return &output, nil
 }
+
+func (kern *Kernel) GetUserHomeDir(ctx context.Context, input *api.GetUserHomeDirInput) (*api.GetUserHomeDirOutput, error) {
+	path, err := os.UserHomeDir()
+	if err != nil {
+		return &api.GetUserHomeDirOutput{}, errors.New("could not get user home directory")
```
brandonbloom, line 207:
```suggestion
		return &api.GetUserHomeDirOutput{}, fmt.Errorf("getting user home directory: %w", err)
```

brandonbloom (approved):

jwmza:
@BenElgar does this / could this also return the parent directory name and full path when I call `readDir`, similar to `ls -a`?

i.e. it'd be nice to get
```
{
  parent: Entry,
  entries: Entry[]
}
```
for purposes of not splitting the path logic between backend and frontend.

internal/core/server/kernel.go
BenElgar: :+1: ah cheers!

BenElgar:
@jwmza I don't entirely understand the use case but I added that to the interface :)
[`a7c859d` (#342)](https://github.com/deref/exo/pull/342/commits/a7c859dcdb0b2dca9f0fdfb3586c645bb04a76c4)

jwmza:
> I don't entirely understand the use case but I added that to the interface :)

Thanks @BenElgar  -  basically the idea is to give the user the ability to click either a child directory or a `..` button to go into the parent directory, i.e. the following GUI:

```js
Root:
/path/to/dir

[ ..     ]      // Clicking does same as "cd .."
[ child1 ]      // Clicking does same as "cd child1"
[ child2 ]      // Clicking does same as "cd child2"
[ child3 ]      // Clicking does same as "cd child3"

```

Therefore, the underlying data takes the current directory as an input and returns both all the children as well as the parent path (parent of the current directory; not the current directory itself)

BenElgar:
Ah, sorry, I misunderstood entirely. That makes more sense. Done that here: https://github.com/deref/exo/pull/349/files

Issue #341, opened 2021-09-21
Create project templating backend


BenElgar:
Should provide the following interface for creating a project:
```
api.kernel.createProject(path: string, templateName?: string): Promise<string>
```
where the return result is a workspace ID.

And the following for listing available templates:
```
api.kernel.listTemplates(): Promise<string[]>;
```
where the return result is a list of template names.

This command should:
1. fail if `path` exists
2. create `path` as a directory
3. If a template was provided:
	a. copy the templated files into the directory
	b. insert a relevant manifest file
5. run the `create-workspace` command
6. return the workspace ID

brandonbloom:
I'd like to propose a few tweaks to this interface.

1) In `createProject`: specify templates by URL, not by name. This way _external_ templates can be supported someday. This way the template ID is just the path to the root of the directory on github or something like that.
2) Extend the return value of `listTemplates` to a richer structure, so that we can add metadata. For example, the template picker UI probably wants a display name:

```typescript
interface TemplateDescription {
  url: string;
  displayName: string;
}
```

PR #343, opened 2021-09-21
Add templating backend

BenElgar:


BenElgar:
@jwmza the implementation might change but it currently correctly implements the interface specified in the linked issue.

jwmza (commented):
This looks good, however I think we ought to have an `empty-project` template in addition to, in this case, the railway ones.

Also, will this create a new directory (recursively?) if the root doesn't exist?

In other words, if my `~/john/dev` directory is empty and I call the API function with `root: "~/john/dev/myorg/projectname"`, will the directories `myorg` and `projectname` be created and the project files stored there?

EDIT: Not sure about relative path but same question stands with `~` replaced by `/home/`

brandonbloom:
Added some design feedback to the #341 thread.

jwmza:
@BenElgar no need to support `~` as the directory picker will only allow you to navigate between paths the UI provides, you won't be able to type the whole path however you want.

What I _would_ like to support is naming your new project e.g. `acme/search/prototype` and having all three directories created as a result.

In this way, the naming convention for projects in the create project flow would be:
- Alphanumerics,
- joined optionally by dashes,
- joined optionally by slashes,
- such that they create paths like `my/path-to/somewhere`

jwmza:
> can you do a mockup or visual of the file browser you're planning to build so we make sure we have all the functionality we need?

There's already a good one in #337 , it really just needs some styling.

image

I don't think showing non-directory files or intermediate directory creation is necessary at the moment. Could add them, but I don't see it as adding value.

The core functionality is easy, simple graphical navigation; project naming; and submit + error handling.

jwmza (approved):
I haven't fully reviewed the Go code so I would suggest @brandonbloom have a look there; but it looks good to me and the `.ts` API looks perfect.

@BenElgar I'm assuming the "Empty project" 'template' I asked about is achieved by calling `api.kernel.createProject` with no second parameter?

BenElgar:
> I'm assuming the "Empty project" 'template' I asked about is achieved by calling api.kernel.createProject with no second parameter?

@jwmza yup!

BenElgar:
I'm going to merge this, just to make collaboration easier. Very happy to address PR comments in follow up PRs.

brandonbloom:
I'll find some time to review this afternoon.

internal/core/server/kernel.go
```diff
@@ -28,6 +30,45 @@ type Kernel struct {
...
+		}
+	}
+
+	err := os.Mkdir(projectDir, 0750)
+	if err != nil {
+		return &api.CreateProjectOutput{}, fmt.Errorf("making project dir: %w", err)
+	}
+
+	if templateDir != "" {
+		copy.Copy(templateDir, projectDir, copy.Options{
```
brandonbloom, line 58: Why do we need to uncompress in to a temporary directory and then copy in to the project directory? Why not directly uncompress in to the project directory?

internal/core/server/kernel.go
```diff
@@ -28,6 +30,45 @@ type Kernel struct {
...
+			OnSymlink: func(string) copy.SymlinkAction {
+				return copy.Deep
+			},
+		})
+	}
+
+	result, err := kern.CreateWorkspace(ctx, &api.CreateWorkspaceInput{Root: projectDir})
+	if err != nil {
+		return &api.CreateProjectOutput{}, fmt.Errorf("creating workspace: %w", err)
```
brandonbloom, line 67:
```
return nil, fmt.Errorf("creating workspace: %w", err)
```

internal/core/server/kernel.go
```diff
@@ -28,6 +30,45 @@ type Kernel struct {
...
 }
 
+func (kern *Kernel) DescribeTemplates(ctx context.Context, input *api.DescribeTemplatesInput) (*api.DescribeTemplatesOutput, error) {
+	return &api.DescribeTemplatesOutput{Templates: template.GetTemplateDescriptions()}, nil
+}
+
+func (kern *Kernel) CreateProject(ctx context.Context, input *api.CreateProjectInput) (*api.CreateProjectOutput, error) {
+	projectDir := input.Root
+	if !filepath.IsAbs(projectDir) {
+		return &api.CreateProjectOutput{}, errors.New("path must be absolute")
```
brandonbloom, line 40: `return nil, errors.New("path must be absolute")`

internal/core/server/kernel.go
```diff
@@ -28,6 +30,45 @@ type Kernel struct {
...
+	}
+
+	var templateDir string
+	if input.TemplateUrl != nil {
+		var err error
+		templateDir, err = template.GetTemplateFiles(ctx, *input.TemplateUrl)
+		if err != nil {
+			return &api.CreateProjectOutput{}, fmt.Errorf("getting template files: %w", err)
```
brandonbloom, line 48: `return nil, ...`

internal/core/server/kernel.go
```diff
@@ -28,6 +30,45 @@ type Kernel struct {
...
+}
+
+func (kern *Kernel) CreateProject(ctx context.Context, input *api.CreateProjectInput) (*api.CreateProjectOutput, error) {
+	projectDir := input.Root
+	if !filepath.IsAbs(projectDir) {
+		return &api.CreateProjectOutput{}, errors.New("path must be absolute")
+	}
+
+	var templateDir string
+	if input.TemplateUrl != nil {
```
brandonbloom, line 44: Empty string is an invalid URL, so might as well enbrace the zero-value idiom and change the type to just be string instead of `*string` then check against `""` here.

internal/core/server/kernel.go
```diff
@@ -28,6 +30,45 @@ type Kernel struct {
...
+
+	err := os.Mkdir(projectDir, 0750)
+	if err != nil {
+		return &api.CreateProjectOutput{}, fmt.Errorf("making project dir: %w", err)
+	}
+
+	if templateDir != "" {
+		copy.Copy(templateDir, projectDir, copy.Options{
+			OnSymlink: func(string) copy.SymlinkAction {
+				return copy.Deep
```
brandonbloom, line 60: Why Deep? Seems like `copy.Shallow` is more appropriate.

internal/template/template.go
```diff
@@ -0,0 +1,51 @@
...
+	err = uncompress(resp.Body, dir)
+	if err != nil {
+		return "", fmt.Errorf("untarring template: %w", err)
+	}
+	return filepath.Join(dir, archiveFolderName), nil
+}
+
+// MakeTemplateFiles takes a directory of files to template and outputs the
+// resultant template into the given directory.
+func MakeTemplateFiles(ctx context.Context, inputDir, outputDir string) error {
```
brandonbloom, line 39: Is this actually make-template-_file_ singular?

internal/template/template.go
```diff
@@ -0,0 +1,51 @@
...
+	tarFile := filepath.Join(outputDir, tarName)
+	f, err := os.Create(tarFile)
+	if err != nil {
+		return fmt.Errorf("failed to create file %q: %w", tarFile, err)
```
brandonbloom, line 43: `creating tar file: %w`

cmd/gentemplates/main.go
```diff
@@ -0,0 +1,83 @@
...
+			templateDir := path.Join(examplesPath, templateName)
+
+			tmplOutDir := path.Join(outDir, templateName)
+			if err := os.Mkdir(tmplOutDir, 0750); err != nil {
+				return fmt.Errorf("creating template dir: %w", err)
+			}
+
+			eg.Go(func() error {
+				if err := template.MakeTemplateFiles(ctx, templateDir, tmplOutDir); err != nil {
+					return fmt.Errorf("making template files: %w", err)
```
brandonbloom, line 75: this error doesn't say _which_ template failed. Since they happen in parallel, you're not going to be able to tell, should you encounter this error. Maybe consider something like `making %q template files: %w", templateDir, err`

cmd/gentemplates/main.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	examplesPath := path.Join(dir, "examples")
+	entries, err := ioutil.ReadDir(examplesPath)
+	if err != nil {
+		return fmt.Errorf("reading examples dir: %w", err)
+	}
+
+	ctx := context.Background()
+	var eg errgroup.Group
```
brandonbloom, line 58: Why do this in parallel? I'd expect this code to be IO bound and parallelism seems like it wouldn't help at all.

brandonbloom:
OK, pretty good. Left quite a few comments, mostly small things.

script/update-templates.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+source "$( dirname "${BASH_SOURCE[0]}" )/.include"
+
+cd "$ROOT_DIR"
+aws s3 sync "$(go run ./cmd/gentemplates/main.go)" s3://exo-starter-templates
```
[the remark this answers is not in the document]
BenElgar: So I was trying to avoid having people download the full set of starters when creating a new project. The full set of starters is about 12 megabytes which would take about 12 seconds to download on a slow internet connection (8Mbps). Downloading a compressed directory means it's at most ~600KB to download.

Happy to revert it to just use git if you think that's better.

internal/template/tar.go
```diff
@@ -0,0 +1,126 @@
+package template
```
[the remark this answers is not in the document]
BenElgar: Yes it's exactly that. I couldn't immediately find a package that was designed for external use that provided this interface.

internal/template/tar.go
```diff
@@ -0,0 +1,126 @@
...
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// From https://gist.github.com/mimoo/25fc9716e0f1353791f5908f94d6e726
+
+func compress(src string, buf io.Writer) error {
```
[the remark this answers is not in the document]
BenElgar: :+1: This just came from the linked implementation. Will change.

internal/template/template.go
BenElgar: My thinking was that the template structure is a directory. It so happens that right now the directory only contains one file (a `files.tar.gz`) but the expectation is that it will contain more for metadata like icons or descriptions of the template.

cmd/gentemplates/main.go
BenElgar: It made it consistently _slightly_ faster for me but so slight as to not be worth including. Will remove.

internal/core/server/kernel.go
BenElgar: I changed it from a copy to a move but the same question still applies. I believe it's conventional for a tarball of a directory to contain the directory itself at the top level. That means that we need to unwrap the containing directory when extracting the tarball. We could do this in the extraction logic but it seemed simpler and clearer to leave that logic untouched and unwrap it using a temporary directory. Happy to change if you think it would be better.

internal/core/server/kernel.go
BenElgar: From the Go perspective that seems reasonable — but do we want that to be true at the API interface? That feels like leaking the implementation detail of the language. Nbd either way.

internal/core/server/kernel.go
BenElgar: No longer relevant since I changed this to a move — but you're probably right.

BenElgar:
Addressed feedback here: https://github.com/deref/exo/pull/355

script/update-templates.sh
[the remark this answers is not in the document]
brandonbloom: No, this makes sense. Thanks for explaining.

internal/template/tar.go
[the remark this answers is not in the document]
brandonbloom: That implementation has some other problems, as i've copy/pasted it before, but I forget what they are :)

internal/template/tar.go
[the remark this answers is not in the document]
brandonbloom: ok, if we find we need it again, we can move to tarutil or whatever & eventually promote it out

internal/core/server/kernel.go
brandonbloom: i've seen tarballs both way, but yeah, it's usually nice to have one top-level directory so that you don't wind pu with a ton of files mixed in with your other stuff when you untar

internal/core/server/kernel.go
brandonbloom: I'm doing some work on formalizing JOSH & going to replace the IDL format with something JSON-based & try to define the schema such that we can generate sensible types for both Go and TypeScript. Stay tuned for that. Very early work on it here: https://github.com/deref/josh

internal/template/template.go
brandonbloom: gotcha - we should probably document the template format somewhere. maybe in a README or `go.doc` file or something. A json schema or typescript signature for the metadata is probably a good idea once we get there.

cmd/gentemplates/main.go
brandonbloom: probably depends on your hard disk & the number of files in parallel - i worry if that number gets large, it just thrashes

PR #356, opened 2021-09-23
Improve directory browser

jwmza:
Adds
- Home button
- Search/filter

image


jwmza:
I would also like to stop flashing only a spinner when the directory changes and the promise is being awaited, instead, I think we should show the result of the old promise until the new one is ready, maybe with a spinner as well but currently it is causing weird visual flashes and the submit button is jumping around a lot.

gui/src/components/DirectoryBrowser.svelte
```diff
@@ -0,0 +1,124 @@
...
+  {:else}
+    <button disabled> <LeftUpSVG /> </button>
+  {/if}
+
+  {#if dir.directory.path !== homePath}
+    <button title="Home directory" on:click={() => handleClick(homePath)}>
+      <HomeSVG />
+    </button>
+  {:else}
+    <button disabled> <HomeSVG /> </button>
```
brandonbloom, line 30: Could you do `disabled={dir.directory.path === homePath}`, this way the title is always hooked up & there is less code?

gui/src/components/DirectoryBrowser.svelte
```diff
@@ -0,0 +1,124 @@
...
+<div class="toolbar">
+  {#if dir.parent !== null}
+    <button
+      title="Parent directory"
+      on:click={() => handleClick(String(dir.parent?.path))}
+    >
+      <LeftUpSVG />
+    </button>
+  {:else}
+    <button disabled> <LeftUpSVG /> </button>
```
brandonbloom, line 22: see comment below about disabled buttons

gui/src/components/DirectoryBrowser.svelte
```diff
@@ -0,0 +1,124 @@
+<script lang="ts">
+  import type { ReadDirResult } from '../lib/api';
+  import HomeSVG from './mono/HomeSVG.svelte';
+  import LeftUpSVG from './mono/LeftUpSVG.svelte';
+  import Textbox from './Textbox.svelte';
+
+  export let dir: ReadDirResult;
+  export let homePath: string;
+  export let handleClick: (path: string) => void;
```
brandonbloom, line 9: Instead of passing functions in as properties like this, you could use an event dispatcher. See the tutorial: https://svelte.dev/tutorial/component-events

However, this seems strange in general. Why are you passing a function or using events instead of using binding? Rather than `handleClick={setWorkingDirectory}`, you would do `bind:path={workingDirectory}` from the consuming component. Then inside this component, you could do `path = whatever`

brandonbloom (approved):
LGTM. I'd take a moment to improve your Svelte idioms though :) See comments.

gui/src/components/DirectoryBrowser.svelte
jwmza: This is because `setWorkingDirectory` also updates the promise, but yeah, I want to redo mostly the whole page's promise handling as noted above.

gui/src/components/DirectoryBrowser.svelte
```diff
@@ -121,6 +121,6 @@
 
   button:focus,
   button:focus-within {
-    background: var(--grey-d-color);
+    box-shadow: var(--shadow-focus) !important;
```
brandonbloom, line 124: important? ಠ_ಠ 

:)

jwmza: Otherwise I'd need to put `:not(:focus):not(:focus-within)` on the hover style

jwmza:
I'm just gonna merge this and then follow up w/ the improvements mentioned above soon. Likely tonight/tomorrow but I will probably prioritize improved component CRUD GUI (cards, dropdown) first.

gui/src/pages/NewProjectConfig.svelte
```diff
@@ -88,24 +91,11 @@
...
-          <div class="directories">
-            {#each dir.entries
-              .filter((x) => x.isDirectory)
-              .sort((x, y) => x.name.localeCompare(y.name)) as entry}
-              <button on:click={() => setWorkingDirectory(entry.path)}
-                >{entry.name}</button
-              >
-            {/each}
-          </div>
+          <DirectoryBrowser
```
brandonbloom, line 94: yes, more smaller components all the time :)

Also, I still strongly recommend you start making more aggressive use of Storybook :)

PR #357, opened 2021-09-23
Add token authentication

BenElgar:
This adds token authentication for each client to the server. There is a `token` file that can contain one or more tokens (to allow for rotation). The token can either be specified in the `Authorization` header as a bearer token or in a `token` cookie.

Right now the user experience isn't the best when the token isn't set so that should be fixed before merging. Otherwise everything works and should be ready for review.


BenElgar:
@jwmza Do you reckon you could put together a "Not authenticated. Please run `exo gui` to continue." page?

internal/config/config.go
```diff
@@ -95,6 +96,9 @@ func setDefaults(cfg *Config) {
 	if cfg.RunStateFile == "" {
 		cfg.RunStateFile = filepath.Join(cfg.RunDir, "exod.json")
 	}
+	if cfg.TokenFile == "" {
```
brandonbloom, line 99: If it can have multiple tokens, should it be called "TokensFile" (plural)?

I'm not clear on how rotation will work & if you need token metadata. Maybe we should practice YAGNI for that & just support one token at a time, or actually think through what is needed for rotation.

internal/core/server/handler.go
```diff
@@ -21,8 +25,26 @@ type Config struct {
...
+		return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+			token := strings.TrimPrefix(req.Header.Get("Authorization"), "Bearer ")
+			cookie, err := req.Cookie("token")
+			if token == "" && err == nil {
+				token = cookie.Value
+			}
+
+			authed := tokenClient.CheckToken(token)
+			if !authed {
+				httputil.WriteError(w, req, errutil.NewHTTPError(http.StatusUnauthorized, "Bad or no token"))
```
brandonbloom, line 39: nit: error messages typically are not complete sentences and therefore should not be capitalized.

internal/token/token.go
```diff
@@ -0,0 +1,96 @@
...
+import (
+	"bufio"
+	"crypto/rand"
+	"encoding/hex"
+	"fmt"
+	"os"
+	"strings"
+)
+
+type TokenClient interface {
```
brandonbloom, line 12: This interface and the way it is being used is kinda strange. You've got an interface, implying that there are potentially multiple implementations of it, but then various key codepaths call `NewFileTokenClient` internally. Also, `NewFileTokenClient` can actually fail.

A better approach would be to make the `GetToken/CheckToken` methods potentially return errors, and then drop the constructor for the `FileTokenClient` altogether. Instead, you'd initialize it like `&FileTokenClient{Path: path}` and any code that needs a client would have one passed in. This would delay the file check until `GetToken` is called, so that if there is a problem with that, other code paths won't fail prematurely.

internal/token/token.go
```diff
@@ -0,0 +1,96 @@
...
+func genToken() (string, error) {
+	buff := make([]byte, 20)
+	if _, err := rand.Read(buff); err != nil {
+		return "", fmt.Errorf("getting randomness: %w", err)
+	}
+	return hex.EncodeToString(buff), nil
+}
+
+func EnsureTokenFile(path string) (*fileTokenClient, error) {
+	tokenFile, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0600)
```
brandonbloom, line 73: Most of the code in this function can be replaced by `ioutil.WriteFile`

internal/token/token.go
```diff
@@ -0,0 +1,96 @@
...
+
+var _ TokenClient = &fileTokenClient{}
+
+func NewFileTokenClient(path string) (*fileTokenClient, error) {
+	tokenFile, err := os.Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("reading token file: %w", err)
+	}
+
+	scanner := bufio.NewScanner(tokenFile)
```
brandonbloom, line 30: Since these files are expected to be small, I might have just done `ioutil.ReadFile` then `Split`, but scanner is good too.

brandonbloom:
Pretty good. Just beating up on you for Go idioms etc for now :)

internal/core/server/handler.go
BenElgar: So I wasn't sure how best to do this for messages that are going to show up directly in the UI. There it looks much better capitalised and possibly punctuated.

cmd/exo/gui.go
```diff
@@ -43,6 +44,16 @@ If the current directory is part of a workspace, navigates to it.`,
 			endpoint = routes.WorkspaceURL(*output.ID)
 		}
 
+		// Add a token to auth.
```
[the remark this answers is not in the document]
BenElgar: You rather persuaded me that having all URLs having the token in them was a bad idea. But I'm happy to be unpersuaded.

internal/core/server/handler.go
brandonbloom: Ah, good point. We need to think through our error strategy more completely. What we probably want is some stable error code that we can potentially use for localization someday.

cmd/exo/gui.go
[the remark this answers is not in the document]
brandonbloom: heh, ok - yeah i forgot what we decided. i was replying from bed in the morning :)

let's stick with it as is - i think it's fine if the links don't work until you do `exo gui` at least one time. we'll see how it feels

gui/src/App.svelte
```diff
@@ -3,6 +3,14 @@
...
   import Offline from './pages/Offline.svelte';
   import { isRunning } from './lib/global/server-status';
+
+  const params = new URLSearchParams(window.location.search);
+  const urlToken = params.get('token');
+  if (urlToken) {
+    document.cookie = `token=${urlToken}`;
+    params.delete('token');
+  }
+  window.location.search = params.toString();
```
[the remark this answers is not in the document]
BenElgar: Did it but now irrelevant since this is all gone.

internal/config/config.go
```diff
@@ -95,6 +96,9 @@ func setDefaults(cfg *Config) {
...
+		cfg.TokenFile = filepath.Join(cfg.HomeDir, "token")
```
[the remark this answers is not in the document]
BenElgar: :+1:

internal/token/token.go
BenElgar: My thinking was that if I want to iterate through each one and do `strings.TrimSpace` it doesn't really win me much.

BenElgar:
Note the important change whereby the token setting is taken out of the client and moved to the server. That makes the redirect experience much smoother and improves resistance to XSS.

internal/config/config.go
BenElgar: Renamed. I don't think the rotation story matters much right now. I added a very simple implementation of it just because it's been my experience that it's one of the rare features that if you only think about it only once you need it, it's too late. But I don't think that any of that is relevant to this very simple use case. Happy to remove entirely if you'd prefer.

internal/token/token.go
BenElgar: I rather thought that the fact that the constructor failed early and that the token file was read only once was a feature — but perhaps that's not the Go way. Will change.

BenElgar:
Fixed the user experience on the front end when not authenticated

internal/token/token.go
brandonbloom: So I won't claim to be the arbiter of the one true "Go way".

My experience though has been that the fewer constructors/factory-functions, the better. Delaying errors makes things more dynamic, lazy-loaded, etc. too. So in this case, we won't fail on startup for like `exo help` or `exo version` or similar when we load config and try to create a client, but something goes wrong!

brandonbloom (approved):
UX changes seem good too. Thanks!

brandonbloom:
@BenElgar on latest main, `dexo logs` fails:

```
~/Projects/deref/exo (main) $ dexo logs
getting token client: reading token file: reading token file: open /Users/brandonbloom/Projects/deref/exo/.dev/var/token: no such file or directory
```

PR #358, opened 2021-09-23
Sqlite logs

### LINKED ISSUE #336: [BUG] `.exo` directory growing very large in size
**Describe the bug**
My production `.exo/var` directory has inflated to >2 GB in size, ~99% of which is in `logs` 

**Expected behavior**
Some sort of reasonable garbage collection to keep the storage used small. Ideally it would be nice to have a user configurable storage limit so that users could decide whether they want unlimited history, or any byte amount.

**Screenshots**
```
~/.exo/var/logs $ ls -l -h -a
total 424K
drwx------ 2 jwmza jwmza 4.0K Sep 15 13:06 .
drwx------ 4 jwmza jwmza 4.0K Sep 17 18:28 ..
-rw-r--r-- 1 jwmza jwmza 303K Sep 14 12:03 000012.sst
-rw-r--r-- 1 jwmza jwmza 7.2K Sep 14 15:59 000013.sst
-rw-r--r-- 1 jwmza jwmza 2.6K Sep 15 13:06 000014.sst
-rw-r--r-- 1 jwmza jwmza   20 Sep 15 13:06 000018.vlog
-rw-r--r-- 1 jwmza jwmza 2.0G Sep 15 13:06 000019.vlog      << This one seems to be the problem
-rw-r--r-- 1 jwmza jwmza 128M Sep 17 18:28 00013.mem
-rw-r--r-- 1 jwmza jwmza 1.0M Aug 19 12:55 DISCARD
-rw------- 1 jwmza jwmza   28 Aug 19 12:55 KEYREGISTRY
-rw-r--r-- 1 jwmza jwmza    4 Sep 15 13:06 LOCK
-rw------- 1 jwmza jwmza  282 Sep 15 13:06 MANIFEST
```

brandonbloom:
- Replace BadgerDB with Sqlite for log storage. This gives us less control, but should be efficient enough and is much easier to work with.
- Rename "log collector" to "event store", and other cleanups.

Fixes #336


brandonbloom:
@BenElgar this is ready for review - but i still need to do a pretty thorough test pass over it.

I'm also tempted to make a tweak to the event table schema before that would require a migration. I'm thinking about adding a json-encoded metadata field of some kind, so that we can add extra data to events. Main use case I'm thinking about is annotating "system events" somehow, so we can display them specially in the GUI or CLI. - What do you think?

internal/gensym/ulid.go
```diff
@@ -24,14 +24,14 @@ func NewULIDGenerator(ctx context.Context) *ULIDGenerator {
 	}
 }
 
-func (gen *ULIDGenerator) NextID(ctx context.Context) ([]byte, error) {
+func (gen *ULIDGenerator) NextID(ctx context.Context) (ulid.ULID, error) {
 	// The math/rand generator is not thread-safe, so we have to guard access with a mutex.
 	gen.mu.Lock()
 	defer gen.mu.Unlock()
 	ts := ulid.Timestamp(chrono.Now(ctx))
 	id, err := ulid.New(ts, gen.entropy)
```
BenElgar, line 32: If not adding additional context to error why not just
```suggestion
	return ulid.New(ts, gen.entropy)
```

go.mod
```diff
@@ -20,9 +20,11 @@ require (
 	github.com/gofrs/flock v0.8.1
 	github.com/hashicorp/hcl/v2 v2.10.0
 	github.com/influxdata/go-syslog/v3 v3.0.0
+	github.com/jmoiron/sqlx v1.2.1-0.20190826204134-d7d95172beb5
 	github.com/joho/godotenv v1.3.0
 	github.com/lucasb-eyer/go-colorful v1.2.0
 	github.com/mattn/go-isatty v0.0.13
+	github.com/mattn/go-sqlite3 v1.9.0
```
BenElgar, line 27: This version of go-sqlite is affected by this bug that's affecting me (in a very minor way):
https://github.com/mattn/go-sqlite3/issues/803

It's fixed in 1.14
```suggestion
	github.com/mattn/go-sqlite3 v1.14.8
```

BenElgar (approved):
All looks good to me! The JSON metadata notion for system events also seems like a good idea.

internal/gensym/ulid.go
brandonbloom: Good catch. This was slightly different code before and I just didn't do the cleanup.

brandonbloom:
OK, @BenElgar fixed your feedback & fixed the UI callsites (d'oh!)

Going to merge this prior to the metadata column, as that needs some more thought. Can deal with the migration later, or just ignore it if this release never goes out like this.

PR #361, opened 2021-09-24
Guard against DNS rebinding

BenElgar:
This guards against DNS rebinding by checking the Host is one that we are expecting to serve for.

### DIFF
--- a/internal/exod/main.go
+++ b/internal/exod/main.go
@@ -2,6 +2,7 @@ package exod
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


internal/exod/main.go
```diff
@@ -42,6 +44,20 @@ func Main(ctx context.Context) {
 	RunServer(ctx, cmd.Flags)
 }
 
+var makeHostHeaderMiddleware = func(validHosts []string) func(next http.Handler) http.Handler {
```
brandonbloom, line 47: This is a case where I think it would be much more Go-idiomatic to use a struct:

```go
type HostAllowListHandler struct {
  Hosts []string
  Next http.Handler
}

func (h *HostAllowListHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
  ...
}
```

The result is shallower nesting, easier debugging (structs are transparent when printed, closures are opaque), etc.

I might also move this to the httputil package just so that it's not cluttering up reading of exod/main.

brandonbloom (approved):
LGTM, idioms aside

internal/exod/main.go
BenElgar: Yeah, fair, that's much less gross.

Issue #347, opened 2021-09-21
[FEATURE] clear log scrollback


brandonbloom:
Sometimes, you want a clear/fresh slate in your log viewer so that you can clearly delineate new log messages quickly and visually. Typical terminals support clearing scroll back, often used for this purpose.

we should have a button in the GUI for this, and support typical hotkeys: cmd+k, ctrl+l

i suspect we could implement this by explicitly adding some kind of system event to the log & then handling the rendering of that specially in the GUI

reported via @finnigantime

PR #362, opened 2021-09-24
Put terminal in to "raw mode" when tailing logs.

brandonbloom:
In particular, this implements the CLI side of #347: handling of ^l.
cmd+k (for MacOS) already worked and remains so.

### DIFF
--- a/cmd/exo/logs.go
+++ b/cmd/exo/logs.go
@@ -80,7 +80,7 @@ func runTailLogsReader(ctx context.Context, cancel func()) error {
 		switch press.Key {
 		// Clear screen.
 		case terminal.KeyCtrlL:
-			fmt.Print("\033[2J")
+			fmt.Print("\033[2J\033[1;1H")
 
 		// Quit.
 		case terminal.KeyCtrlC, 'q':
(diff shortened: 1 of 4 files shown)


brandonbloom:
This was substantially more work than I expected!

@jwmza could you please test this on windows?

@BenElgar please test on linux.

Thanks :)

jwmza:
Yeah! Will do

BenElgar (commented):
The implementation seems reasonable but I'm not sure it's behaving as expected. Clearing the screen doesn't reset the cursor position as I would expect.

https://user-images.githubusercontent.com/3443024/134899532-bd5e15c6-6e63-4a85-911d-7bd6dea57119.mp4

jwmza:
@brandonbloom @BenElgar what is `^I` supposed to do, I tried it in a regular (non exo) program and it just seems to indent the logs(?)

BenElgar:
@jwmza That's a lowercase el (`l`). `^l` is the Emacs shortcut for clearing the screen. A lot of programs have copied that shortcut, along with a bunch of other Emacs shortcuts. The most relevant is readline[0], which is what bash uses to read input. 

[0] https://en.wikipedia.org/wiki/GNU_Readline

jwmza:
Thanks @BenElgar

brandonbloom:
Good catch @BenElgar! Fixed by adding a reset command to position (1,1).

PR #363, opened 2021-09-24
Untangle format logs

brandonbloom:
Styling logic for log entries was spread across ~3 files and involved an extra unnecessary type transformation. This untangles that and adds a story for the pure Logs component.


jwmza (approved):
LGTM just looking at the code - but I'll come back and test these changes out later (probably tomorrow).

Much cleaner this way. Thanks!

PR #365, opened 2021-09-24
FORCE_COLOR & fix default environment for .env

### LINKED ISSUE #345: [FEATURE] Force colors where possible
A key motivation for #344 (PTY support) is to convince programs to print colors in logs. Full PTY support is a tall order, so until then, maybe we should do some buildpack-like magic to determine which environment variables to set to encourage programs to override their PTY detection.

For example, `FORCE_COLOR=1` for the npm chalk package used by next.js. See https://github.com/vercel/next.js/issues/11836#issuecomment-612580547

### LINKED ISSUE #347: [FEATURE] clear log scrollback

brandonbloom:
Closes #345

### DIFF
--- a/internal/core/server/workspace.go
+++ b/internal/core/server/workspace.go
@@ -411,29 +411,43 @@ func (ws *Workspace) DescribeEnvironment(ctx context.Context, input *api.Describ
 func (ws *Workspace) getEnvironment(ctx context.Context) (map[string]string, error) {
 	envPath, err := ws.resolveWorkspacePath(ctx, ".env")
 	if err != nil {
-		return nil, fmt.Errorf("could not resolve path to .env file: %w", err)
+		return nil, fmt.Errorf("resolving env file path: %w", err)
 	}
 
-	_, err = os.Stat(envPath)
+	env := map[string]string{}
+
+	// Encourage programs to log with colors enabled.  The closest thing to a
+	// standard for this is <https://bixense.com/clicolors/#bug-reports>, but
+	// support is spotty. This may grow if there are other popular enviornment
+	// variables to include. If we grow PTY support, this may become unnecessary.
+	env["CLICOLOR"] = "1"
+	env["CLICOLOR_FORCE"] = "1"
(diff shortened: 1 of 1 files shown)


internal/core/server/workspace.go
```diff
@@ -411,29 +411,43 @@ func (ws *Workspace) DescribeEnvironment(ctx context.Context, input *api.Describ
...
-	_, err = os.Stat(envPath)
+	env := map[string]string{}
+
+	// Encourage programs to log with colors enabled.  The closest thing to a
+	// standard for this is <https://bixense.com/clicolors/#bug-reports>, but
+	// support is spotty. This may grow if there are other popular enviornment
+	// variables to include. If we grow PTY support, this may become unnecessary.
+	env["CLICOLOR"] = "1"
+	env["CLICOLOR_FORCE"] = "1"
+	env["FORCE_COLOR"] = "3" // https://github.com/chalk/chalk/tree/9d5b9a133c3f8aa9f24de283660de3f732964aaa#supportscolor
```
BenElgar, line 425: This seems to force 24-bit colour support which, according to this link[0] at least, isn't supported by a number of major terminals, including the built-in MacOS terminal app. I don't know whether those terminals will gracefully downsample, just ignore the colours or violently explode.

[0] https://github.com/termstandard/colors#not-supporting-true-color

BenElgar (approved):
LGTM aside from for comment. I don't understand how this closes #347 though.

brandonbloom:
Whoops! #347 was a typo. Ment to write #345. Fixed commit & description. Thanks for catching that!

internal/core/server/workspace.go
brandonbloom: Sorry for the delayed response here, but I just went and tested chalk directly with FORCE_COLOR set to 0, 1, 2, and 3 and printing colors in my Terminal.app locally. Seems to work? ¯\_(ツ)_/¯ 

If this is a problem on Linux or something, let me know.

PR #367, opened 2021-09-27
Remove all support for CORS

BenElgar:
Given all requests are served from a single server, both in dev and prod, I see no reason for cross-origin resource sharing. I think this is a leftover of when we had a separate frontend server that we were using directly in dev.

@brandonbloom am I missing something?


brandonbloom:
As discussed, I added the CORS support before we had a proxy, so I think it is 100% safe to remove now.

PR #369, opened 2021-09-27
Add basic workspace details view

BenElgar:
image
image

Also fixes truncation for the panel title


gui/src/components/ProcessList.svelte
```diff
@@ -59,6 +60,14 @@
   --panel-padding="0 1rem"
 >
   <div slot="actions">
+    <IconButton
+      tooltip="Workspace details"
+      on:click={() => {
+        router.push(`#/workspaces/${encodeURIComponent(workspaceId)}/info`);
+      }}
+    >
+      <PreferencesSvg />
```
jwmza, line 69: _nit:_ I've set a convention of calling all the SVGs `SomeNameSVG.svelte` and using the fully capitalized `<SomeNameSVG />` for the component instances. Just keeps the names predictable (and identical between filename and component name)

gui/src/pages/WorkspaceInfo.svelte
```diff
@@ -0,0 +1,51 @@
...
+  const workspaceRoute = `/workspaces/${encodeURIComponent(workspaceId)}`;
+</script>
+
+<Layout>
+  <WorkspaceNav {workspaceId} active="Dashboard" slot="navbar" />
+  <CenterFormPanel title="Workspace details" backRoute={workspaceRoute}>
+    {#await workspace.describeSelf()}
+      <Spinner />
+    {:then description}
+      <table>
```
jwmza, line 21: Please wrap generic tables with `<CheckeredTableWrapper />` for nice styling out of the box.

gui/src/pages/WorkspaceInfo.svelte
```diff
@@ -0,0 +1,51 @@
...
+
+  const workspaceId = params.workspace;
+  const workspace = api.workspace(workspaceId);
+  const workspaceRoute = `/workspaces/${encodeURIComponent(workspaceId)}`;
+</script>
+
+<Layout>
+  <WorkspaceNav {workspaceId} active="Dashboard" slot="navbar" />
+  <CenterFormPanel title="Workspace details" backRoute={workspaceRoute}>
+    {#await workspace.describeSelf()}
```
jwmza, line 18: Nice @BenElgar , can I use `workspace.describeSelf()` to avoid the `describeWorkspaces().then( [find workspace with same id] )` I mention in #335 ?

gui/src/routes.ts
```diff
@@ -22,6 +23,7 @@ export default {
   '/new-project': NewProject,
   '/new-project/:starter': NewProjectConfig,
   '/workspaces/:workspace': Workspace,
+  '/workspaces/:workspace/info': WorkspaceInfo,
```
jwmza, line 26: _nit:_ I think I prefer `EntityDetails` and `:entity/details` over `EntityInfo` and `:entity/info` - can we establish some nice naming convention for single-entity detail pages?

gui/src/pages/WorkspaceInfo.svelte
```diff
@@ -0,0 +1,51 @@
...
+          <td>{description.root}</td>
+        </tr>
+      </table>
+    {/await}
+
+    <!-- Databases, Apps, cloud services, etc. -->
+  </CenterFormPanel>
+</Layout>
+
+<style>
```
jwmza, line 41: I would use `<CheckeredTableWrapper />` instead of doing this custom styling, and I would also recommend using the two-column `.label` and value approach used in the process details and environment pages currently instead of putting these `<th>`s above `<td>`s

gui/src/components/ProcessList.svelte
```diff
@@ -59,6 +60,14 @@
   --panel-padding="0 1rem"
 >
   <div slot="actions">
+    <IconButton
```
jwmza, line 63: This is good now, but please note that #335 is going to replace these `ProcessList` action buttons with a single `...` Ellipsis button which opens a dropdown menu for actions like adding a new component, seeing workspace details, etc.

Luckily we didn't overlap on making the actual details page @BenElgar

jwmza (changes requested):
LGTM other than the few things I mentioned - @BenElgar it would be nice if we can quickly polish and merge this soon such that I can add this page and conform the buttons/routing to the new `ProcessList` dropdown menu I'm adding in #335

gui/src/pages/WorkspaceInfo.svelte
jwmza: Whoops, didn't realize this already did two-column. Regardless, please conform this to the way we do it in the two pages mentioned above so that we don't have multiple visually different implementations of the same kind of table. I'll probably make a `KeyValueTable` component like we had in the old console (I think they were called "Descriptions" or "Summaries" per the old design framework?)

jwmza:
#335 is ready to be integrated with this whenever it is ready. Will be some merge conflicts but nothing major.

gui/src/components/Panel.svelte
```diff
@@ -78,6 +78,7 @@
     overflow: hidden;
     text-overflow: ellipsis;
     white-space: nowrap;
+    display: inline-block;
```
jwmza, line 81: Why?

gui/src/components/ProcessList.svelte
```diff
@@ -94,3 +103,10 @@
     </IfEnabled>
   </section>
 </Panel>
+
+<style>
+  div[slot='actions'] {
+    display: flex;
+    flex-direction: row;
```
jwmza, line 110: _nit_: `flex-direction: row` is the default value.

brandonbloom:
Thanks @BenElgar for taking the initiative on this. Sadly, I think you're going to conflict with a bunch of stuff @jwmza is working on!

@jwmza would you mind taking this over from Ben after you land your other process list changes?

jwmza:
Yeah @brandonbloom  I'll merge #335 first and then conform this PR & merge it as well, if @BenElgar doesn't have other plans for it :)

gui/src/components/Panel.svelte
BenElgar: This was necessary to fix the ellipsis overflow for the panel title. Right now it doesn't work for me at all — instead it's just silently truncated.

jwmza:
@BenElgar is there anything remaining missing here in terms of functionality, or can I just polish it and merge without worrying?

BenElgar:
I mean the ability edit the fields would be nice but that's a whole different feature. Otherwise I think it's basically complete.

gui/src/pages/WorkspaceInfo.svelte
BenElgar: I don't see why not.

gui/src/components/Panel.svelte
jwmza: I'm not sure what the underlying problem here was but I think this is the wrong solution - what we want is a `display: block` section that spans all the available space between the left and right action groups.

Could you post a screenshot or something showing the issue you ran into without this change? There may be a better fix.

jwmza:
Edit/update flows for workspaces will come alongside those for components, which is my next big priority. Merging this now, thanks all!

gui/src/components/Panel.svelte
BenElgar: Sure — I leave it up to you to decide the right fix :)
https://user-images.githubusercontent.com/3443024/135058170-d720772e-09e1-415c-b67c-17efc8fff8b5.mp4

jwmza: I see. My first look at the issue makes me think perhaps the solution is `width: calc(100%)`, as strange as that is.

To explain what we _want_ here, the panel header should have its width automatically filling up the width of the panel, and the title portion should take up all the space between both left and right action sections - i.e. `max-content auto max-content` grid column layout.

Apparently `text-overflow: ellipsis` only renders when the width is constrained in absolute units like pixels. However, `calc()` outputs as absolute pixels, so I *think* if we wrap the title in a block with `width: calc(100%)` we can have the `auto` width _and_ the `ellipsis` working properly.

That all being said, the content of this title has been changes so it should usually not overflow anymore as `Workspace af28109rjwefjsdiqq...` has now changed to `my/project/name` where it uses the shortest unique path. Should still overflow sometimes but usually not.

PR #372, opened 2021-09-27
Fix handling of no streams in sqlite code

brandonbloom:


BenElgar:
I'm not necessarily clear on why this is more correct than just returning an error when no streams are requested. What does the value of the cursor represent when no streams are requested?

brandonbloom:
It's not about being more _correct_, it's about being more general. Before the sqlite change, the backend could handle the case when no streams were passed. After, that flexibility broke, which broke a client. We violated postel's law, which meant that clients had to be updated to do conditional logic, etc. Not great. This PR restores the generality to the backend.

Specifically regarding the cursor when there are no streams: The contents of the cursor is opaque. The client expects to get a valid cursor & that cursor should be stable regardless of the set of streams. So if you toggle streams on and off, you don't lose your scroll position. In this implementation, a synthetic event ID qualifies, as it is prefixed by a timestamp & has a per-process monotonic increasing sequence id.

PR #373, opened 2021-09-27
Add noDaemon config.

brandonbloom:
This is largely in service of error prevention when using dexo.

I made the error of using `dexo daemon` or `dexo exit` a few times, but the server is actually expected to be running under the production `exo`'s supervision. This change presents some sensible error messages when in this mode, so that I don't screw up and to force me to dogfood exo itself more while working on exo.


BenElgar:
:+1:

PR #374, opened 2021-09-27
Add dynamic `Icon` component

jwmza:
Adds a single `Icon` component with named `glyph` attribute so icons can be dynamic in data

For example, instead of

```
import IconASVG from "./mono/IconASVG.svelte";
import IconBSVG from "./mono/IconBSVG.svelte";
import IconCSVG from "./mono/IconCSVG.svelte";
...
<IconASVG />
<IconBSVG />
<IconCSVG />
```

you write

```
import Icon from "./Icon.svelte"
...
<Icon glyph="IconA" />
<Icon glyph="IconB" />
<Icon glyph="IconC" />
```

This means we can finally get the best of both worlds by having Svelte component SVG icons which can be dynamically looped over and chosen with data.

### DIFF
--- a/gui/src/components/CheckboxButton.svelte
+++ b/gui/src/components/CheckboxButton.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import CheckSVG from './mono/CheckSVG.svelte';
+  import Icon from './Icon.svelte';
 
   export let active: boolean = false;
   export let tooltip: string = '';
@@ -12,7 +12,7 @@
 >
   {#if active}
     <div>
-      <CheckSVG />
+      <Icon glyph="Check" />
     </div>
   {/if}
 </button>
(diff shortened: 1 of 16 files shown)


jwmza:
Some components will soon take advantage of this, notably `IconButton`.

jwmza:
Also, the "New component" menu can finally be dynamically rendered from data instead of hardcoded.

This also lays the groundwork for adding logos/icons to the new "New project" page where it lists all the template framework.

jwmza:
@BenElgar and @brandonbloom please both take a look at this briefly and be aware of the change so we don't accidentally let something new slip through the cracks and end up using the old system or breaking - I will also check myself though.

brandonbloom (approved):
Seems like a reasonable first step to me.

jwmza:
The next step after this is a fully dynamic `DropdownMenu` component that generalizes the extremely verbose styling and icon imports I've done in the two new dropdown menus in #335

PR #375, opened 2021-09-27
Add dynamic dropdown `ContextMenu` component

jwmza:
No visual changes to the generic component, just adds a dynamic component to create the two dropdown menus I added in #335 

- Also now adds `IconGlyph` type information to help when developing with `Icon`, `IconButton`, and `ContextMenu`. More soon.


jwmza:
This should be good to go, one last thing I might do is set up code gen for the `Icon` file since it's stupid for us to waste our time upkeeping a simple file whose contents is a basic transform of the `ls` output of the SVG icon directory!

PR #376, opened 2021-09-27
disable secure cookie on localhost

brandonbloom:
Fixes gui token auth in Safari. Sounds like @jwmza his this with Brave as well.


BenElgar (commented):
Ah sorry — I looked up what this would do in Chrome and Firefox and found that it was fine but I forgot about Safari.

PR #378, opened 2021-09-28
Workspace destroy in GUI & danger style for dangerous `ContextMenu` actions

jwmza:
As @BenElgar mentioned to me, this invokes a job that might fail or take a while so the current implementation might be a little too optimistic.

I've tested it ~8 times and only 1 time did the home page still contain a button to the deleted workspace - still kinda bad. Perhaps should have some sort of loading thing. Other than that, this is good to go.

image
image


jwmza:
Also, _**arguably**_ you could want a modal pop up confirmation here since destroying a workspace in a single accidental click can fuck up a lot of work and configuration.

That being said, I think "execute then offer to undo" is better than "require multiple clicks and pop up an annoying confirmation dialogue to execute"

gui/src/lib/api.ts
```diff
@@ -359,7 +363,11 @@ export const api = (() => {
         return networks;
       },
 
-      async apply() {
+      async destroySelf(): Promise<void> {
```
brandonbloom, line 366: Why not call this "destroy" to match the underlying method? Ideally, this api.ts file will be code-gen'd in the not-too-distant future. I've been occasionally working on http://github.com/deref/josh-js for that.

brandonbloom (approved):

gui/src/lib/api.ts
jwmza: Done

jwmza:
>Alternatively, since this is a rare action, we should probably move it to a danger section on the workspace details page & leave it out of the context menu altogether.

That was my first consideration before putting it in the context menu, but I actually don't think it's that rare, which is why I decided to put it here - but yeah it would have a confirmation.

jwmza:
Added a rough modal but we will need a more sophisticated system to efficiently handle modals from within any component

image

jwmza:
Gonna merge this as-is and loop around when or before we need more modals in order to set up a global store-based modal system.

PR #381, opened 2021-09-28
Use a single combined log stream for each component.

brandonbloom:
Much simpler now :)

Next step will be to include the workspace itself as a log stream, so we can have system events there. Should be a fast follow-up PR.


brandonbloom:
Closing as subsumed by #383

PR #382, opened 2021-09-28
Add --print flag to `exo gui`

brandonbloom:


jwmza:
Awesome, thanks @brandonbloom :)

PR #383, opened 2021-09-29
Workspace stream

brandonbloom:
image


brandonbloom:
Going to merge this as it has many fixes for main & I'd like to roll forward as I continue to improve this stuff.

@BenElgar & @jwmza please do review, as these are pretty big changes to the way the log streams stuff is organized.

The internals of the server workspace handling logic continues to get uglier and uglier. A major refactor of that will be needed at some point in the not-too-distant future, but for now the primary goal is of this PR is to add these system events so that onboarding is less painful.

BenElgar (commented):
LGTM

gui/src/components/logs/LogRow.svelte
```diff
@@ -1,26 +1,20 @@
...
-
-  const friendlyName = streamToFriendlyName(event.stream);
-
-  // SEE NOTE [LOG_COMPONENTS].
-  const trimLabel = (name: string) => name.replace(/(:err$)|(:out$)/g, '');
+  const componentName = getComponentName(event.stream) ?? event.stream;
 </script>
 
-<tr style={logStyleFromHash(friendlyName)}>
+<tr style={logStyleFromHash(componentName)}>
```
jwmza, line 17: Note: this should be updated in `ProcessList` to align the two color format inputs if they aren't already @brandonbloom , so that the same process has the same color in both panels.

jwmza (commented):
LGTM

gui/src/components/logs/LogRow.svelte
brandonbloom: Nothing actually changed about this here, as best I can tell, so it should remain aligned with the ProcessList. However, I think _both_ of them should be updated to use the component id instead of the component name as the hash input value, so that the colors remain the same if a component is renamed (which is a power we will have soon, when I overhaul the manifest system - which i am working on now).

@jwmza Could you make that change?

brandonbloom: (the stream name is always equal to the component id now)

jwmza: Yeah, will do!

PR #385, opened 2021-09-30
State history

brandonbloom:
This is very much work-in-progress. Just sharing for awareness and potential feedback.

The new interface in this PR supports component state storage independent from specification. This simplifies/isolates the logic for state handling. The new interface supports _historical_ states, which will be useful for debugging. Isolated state storage also means that we may be able to support _remote_ state storage, which is important for stacks shared between multiple users in a deployment scenario.

The intention is that many controller actions should be able to operate entirely on state without component specifications. This means that existing controllers might have to change to copy some additional information into the state to make that possible. This will help with recovering from bad specs (#270).

This PR is safe to land, but isn't necessarily worth it to do so, since the new state storage interface isn't actually in use yet. To avoid complex multi-stage migrations, I'll probably storm forward with the new component storage step as well, before merging or after merging, but before switching over.

To be clear: This is step 1, extracted from a very large refactor.

### DIFF
--- a/internal/compstate/api/store.go
+++ b/internal/compstate/api/store.go
@@ -1,0 +1,59 @@
+// Generated file. DO NOT EDIT.
+
+package api
+
+import (
+	"context"
+	"net/http"
+
+	josh "github.com/deref/exo/internal/josh/server"
+)
+
+type Store interface {
+	SetState(context.Context, *SetStateInput) (*SetStateOutput, error)
+	GetStates(context.Context, *GetStatesInput) (*GetStatesOutput, error)
+}
+
+type SetStateInput struct {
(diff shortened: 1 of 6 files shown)


brandonbloom:
Concepts need to evolve here. This component state store might actually be better as a resources store. The idea is that "component state" is roughly the most-recent/current "resource" assigned to the component. This is usually not an important distinction, but in the future, there may be two simultaneous resources for a component. For example, when deploying a new version to replace an old version with zero downtime.

brandonbloom:
Closing this, as I'm basically going to rework all of the internals. May reuse a bunch of this code at some point, we'll see.

PR #387, opened 2021-09-30
Force controllers to use state in preference to specs.

brandonbloom:
Exceptions are initialize & refresh.

Further attempts to mitigate #270.

### DIFF
--- a/internal/providers/docker/components/container/build.go
+++ b/internal/providers/docker/components/container/build.go
@@ -1,3 +1,5 @@
+// SEE NOTE: [IMAGE_SUBCOMPONENT].
+
 package container
 
 import (
@@ -12,6 +14,7 @@ import (
 	"strings"
 
 	"github.com/deref/exo/internal/core/api"
+	"github.com/deref/exo/internal/providers/docker/components/image"
 	"github.com/deref/exo/internal/task"
 	"github.com/deref/exo/internal/util/pathutil"
 	"github.com/docker/docker/api/types"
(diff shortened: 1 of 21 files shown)


brandonbloom:
@BenElgar I need to do a fair bit of testing on this, but it should be reviewable.

brandonbloom:
OK, did a bunch of testing & this seems good. Going to merge. Post-hoc review welcome.

internal/providers/docker/components/container/build.go
```diff
@@ -22,31 +25,37 @@ import (
 )
 
 func (c *Container) Build(ctx context.Context, input *api.BuildInput) (*api.BuildOutput, error) {
-	if c.canBuild() {
-		if err := c.buildImage(ctx); err != nil {
+	if c.State.Image.Spec == "" {
+		// SEE NOTE: [MIGRATE_CONTAINER_STATE].
+		return nil, errors.New("refresh needed")
```
BenElgar, line 30: When will this refresh happen? Do people need to run the manual `exo refresh` command?

BenElgar (commented):
It all LGTM but I'm not quite sure I understand the point. Isn't the issue that a new spec doesn't appropriately replace an invalid state? How does using the state in preference to the spec help with that?

brandonbloom:
> Isn't the issue that a new spec doesn't appropriately replace an invalid state? How does using the state in preference to the spec help with that?

Prior to this change, every operation – including removing a component – relied on being able to parse the spec. Since specs are user-provided, they are much more likely to have syntax errors and other problems. State being incorrect, on the other hand, is a sign of a bug in the provider/component.

PR #388, opened 2021-09-30
Control message not function

brandonbloom:
Goal is to substantially simplify the internals of the workspace implementation by insisting that components are controlled exclusively by message sends, instead of by arbitrary callback functions. This PR is a first step.

Turns out that this was relatively close to true already, minus some error handling and some create/initialize/dispose/delete weirdness.


brandonbloom:
This is pretty likely to break something, but some cursory testing works. Going to merge, since there are already a lot of high risk things in this release & we need to do a proper test pass.

PR #392, opened 2021-10-04
Delete all component types in GUI

jwmza:
This adds an optional `actions` column to the `ComponentTable`s, allowing for arbitrary actions to be taken with icon buttons for each table row.

I've added delete for each component type, to the "Components", "Networking", and "Storage" tabs.

image

However, I don't think we should combine this table data view with CRUD actions like this - not sure.
I think it's probably better if:

1. the components page goes away, and we expand the list of components in the workspace home page from only runnables to all components.
2. we can either delete or keep the Networking and Storage pages, but it doesn't really add much to have them exist until we can show specific information about networking and storage. For example, if we do have a storage page with a table of volumes and other storage components, we should probably do something like show bars for their relative total storage size or something.

### DIFF
--- a/gui/src/pages/WorkspaceNetworking.svelte
+++ b/gui/src/pages/WorkspaceNetworking.svelte
@@ -30,6 +30,7 @@
           getValue: (network) => network.name,
         },
       ]}
+      actions={[]}
     />
   </Panel>
 </Layout>
--- a/gui/src/pages/WorkspaceNetworking.svelte
+++ b/gui/src/pages/WorkspaceNetworking.svelte
@@ -30,7 +30,16 @@
           getValue: (network) => network.name,
         },
       ]}
-      actions={[]}
+      actions={[
+        {
+          tooltip: 'Delete network',
+          glyph: 'Delete',
+          callback: async (component) => {
+            await workspace.deleteComponent(component.id);
+            window.location.reload();
+          },
+        },
+      ]}
     />
   </Panel>
 </Layout>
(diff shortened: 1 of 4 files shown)


gui/src/pages/WorkspaceNetworking.svelte
```diff
@@ -30,6 +30,16 @@
           getValue: (network) => network.name,
         },
       ]}
+      actions={[
+        {
+          tooltip: 'Delete network',
+          glyph: 'Delete',
+          callback: async (component) => {
+            await workspace.deleteComponent(component.id);
+            window.location.reload();
```
brandonbloom, line 39: Surely we can do better than refreshing the entire page when the data changes...

A few concrete suggestions here:

1) Try to give callbacks a more descriptive name. In many cases, that would be something like `onEventName`. Here, it might be something like `execute`.
2) Event callbacks should _usually_, accept an extensible `event` argument. If you call this `execute`, then you might want to just have it be `context`. This will allow you to have a place you can stash extra information if you need it. See next point.
3) You can pass control objects/functions as context to the callback. So you might have something like: `async execute({ reload }, component) { ... }` which allows the ComponentTable component to expose a non-full-page-refresh.

jwmza: Addressed (1) in #394

brandonbloom:
Still a draft?

jwmza:
I suppose not - but this one is waiting for #394 to merge into it (though I suppose it can also merge beforehand)

PR #393, opened 2021-10-04
Add basic support for exo secrets vault

BenElgar:
Some TODOs remain but this is a working implementation of authing via `exo` and making secrets available in the environment.

### DIFF
--- a/cmd/exo/secrets.go
+++ b/cmd/exo/secrets.go
@@ -1,0 +1,173 @@
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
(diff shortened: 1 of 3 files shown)


brandonbloom:
Thanks! I'll look more closely soon. A really brief look though raises one key question: Does all of this happen in the exo cli? Are there any APIs to expose from the exo kernel or workspace such that the GUI can interact with all this stuff too?

jwmza:
@BenElgar please ping me when the UI for this is stable and I'll follow up with a quick styling pass for e.g. the buttons.

brandonbloom:
@jwmza Unless Ben said otherwise, I think the UI is stable for this PR.

PR #394, opened 2021-10-04
Add wip Edit Component page

jwmza:


jwmza:
I think the right way to handle these URL parameters for human-named entities is to use `slug = id | name-id`, in other words, if your component is named `web` with id `123`, then all of the following will work:
```
/web-123
/123
/something-else-123
```
and the URL will be silently changed to `/web-123` regardless, just like Notion pages work.

brandonbloom:
Regarding the URLs: Let's tackle that as a separate bit of work & can discuss the details of it there. Let's get editing working with just IDs in URLs first.

jwmza:
image

@brandonbloom in cases like this where the `spec` of a component is long & single line, do you think we ought to format it human-readable multiline when editing?

brandonbloom:
> in cases like this where the spec of a component is long & single line, do you think we ought to format it human-readable multiline when editing?

Let's fix that on the backend - I'll change the internals to pretty-print specs. Won't help for existing specs, but will help going forward.

PR #395, opened 2021-10-05
Use "GoReleaser"

brandonbloom:


brandonbloom:
@BenElgar I need to study the release machinery that @kendru put together to make sure I don't screw anything up switching over to this. Then I'll get the Homebrew package working. Once that's good, I'll ask you to investigate the Linux packaging story.

No ETA yet, this is stacked up behind a few things on my queue, so let me know if you get spare cycles for the Linux stuff before I get to the Mac stuff.

Thanks!

PR #401, opened 2021-10-06
Edit & rename components in GUI

jrysana:
This assumes #400 's backend without having it so I'm putting it in a separate PR for a moment

### DIFF
--- a/gui/src/pages/EditComponent.svelte
+++ b/gui/src/pages/EditComponent.svelte
@@ -57,6 +57,7 @@
     example: string;
   }
 
+  // TODO: Extract these to a registry.
   const componentTypes: Record<string, ComponentType> = {
     process: {
       glyph: 'Layers',


jwmza:
I tried using this but it seems broken - can't tell if it's a problem with the Svelte client (I don't think so?) or the backend.

After trying to rename a UNIX process **"web2"** to **"web3"** without modifying the spec:
```
14:53:04    server 2021/10/06 14:53:04 http dsgxkrf0q7fdgn0dzpzkz30s74: error processing request: initializing replacement "web2": unmarshalling spec: invalid character 'w' looking for beginning of value
```

After trying to modify the spec of a Docker network **"mynet5"** without modifying the name:
```
14:54:48    server 2021/10/06 14:54:48 http zreght8rycsr08hdn6g1a0w9x0: error processing request: initializing replacement "mynet5": unmarshalling spec: [1:1] string was used where mapping is expected
14:54:48    server     >  1 | mynet5
14:54:48    server            ^
14:54:57    server 2021/10/06 14:54:57 network to be removed not found: "0ab15964153aab12115e4edf02fc75b5783f3f021946076a313e7f809a501308"
```

brandonbloom:
Yeah, the update-component method is currently broken. Working on a fix.

jwmza:
This seems to be working now with one major exception, which is that components simply do not get renamed

PR #404, opened 2021-10-08
Fix excessive refetching causing unnecessary re-rendering.

brandonbloom:
This occurred because of flagging between 'success' and 'refetching'
states, which was changing the element structure with the extra
slot=refetching wrapper.


jwmza:
I take it this fixes the issue where dropdowns would disappear after a few seconds in the `ProcessList`, due to refreshing data?

brandonbloom:
Yes

PR #410, opened 2021-10-13
Experiment: format directory names with `s/^`(home)`/~/g`

jrysana:
Not sure this is actually worth doing (at least, in this manner) but this changes a path like `/home/john/path/to/project` to `~/path/to/project` which removes a lot of repetitive, likely unimportant, information from some locations (as in the `Home` page).

Having to call the API each time for each formatted directory seems like a bad solution though.

### DIFF
--- a/gui/src/components/FormattedDirectory.svelte
+++ b/gui/src/components/FormattedDirectory.svelte
@@ -1,0 +1,15 @@
+<script lang="ts">
+  import { api } from '../lib/api';
+
+  export let raw: string;
+</script>
+
+<span title={raw}>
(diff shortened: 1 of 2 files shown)


brandonbloom (commented):
It's a nice idea. As with any nice UI feature, it's always harder to do well then you'd first thing :)

I left a variety of comments. I'd encourage you to (A) use Storybook! (B) consider how far you want to take the feedback I provided regarding lifting up the async request.

brandonbloom:
Hey @jwmza, I'm going to close this. Feel free to re-open if you want to put in the extra elbow grease to get this in to shape.

PR #415, opened 2021-10-15
Add (`H`)ome and (`P`)references shortcuts to `Layout`

jrysana:
This adds some convenience shortcuts to navigate to the (`H`)ome and (`P`)references pages.

Quite useful in combination with `Alt + Left` for quickly changing preferences and heading back to your previous page (e.g. testing themes in dev, but will expand to more commonly accessed user preferences as that grows)

We might want to do similar things for pages within workspaces as well.

We might also want to use something like `Ctrl+Shift+P` instead of `P`, but I think it's more convenient this way so I don't see a great reason for that.

---

This also merges `<NavbarRoute />` and `<NavbarButton />` by giving the latter an optional `href` attribute.

### DIFF
--- a/gui/src/lib/actions/shortcut.ts
+++ b/gui/src/lib/actions/shortcut.ts
@@ -19,7 +19,9 @@ export const shortcuts = (node: HTMLElement, params: ShortcutsParams) => {
   const removeHandler = () => window.removeEventListener('keydown', handler),
     setHandler = () => {
       removeHandler();
-      if (!params) return;
+      if (!params) {
+        return;
+      }
       handler = (e) => {
         const matched = params.chords.some((chord) => {
           if (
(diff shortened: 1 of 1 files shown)


jwmza:
Agreed @brandonbloom 

Once we start having useful shortcuts for lots of useful actions, we should have a good readily accessible keyboard help popup probably in the bottom-right ellipsis menu in the left toolbar.

As much as keys like `Ctrl`, `Shift`, `Alt` etc make commands less efficient and approachable, I still think these are generally better than `X then Y` type sequences - at least, looking at my own experience as a user of many services, I have tons of combination shortcuts in my conscious and muscle memory, but I _**never**_ use those sequence ones or bother to learn them.

Maybe this experience is not generalizable to the broader audience though?

jwmza:
> If your tools aren't doing that automatically, should investigate why. There should be a prettierrc iirc.

Was replicating the line from d1d05ed , so it seems neither of us have it in automated formatting.

Fixing manually for now.

PR #421, opened 2021-10-17
Fix Firefox subpixel rendering bug

### LINKED ISSUE #414: [BUG] Buttons have no lids
image

My understanding is that this is to do with inconsistent sub-pixel rendering between browsers. I experience the same on mobile. My proposal would be to ensure that everything has at least a thin solid border and then do 3D effects around it.

jrysana:
Applies a minimal 1px outline to all shadows and rebalance other shadows in order to fix Firefox's subpixel rendering bug.

Fixes #414 

I'm conflicted on whether this should actually be solved this way. Ultimately, this is Mozilla's fault in rendering subpixel shadows incorrectly (there's really no room for debate on this as far as I understand it - if designers intentionally create a shadow with subpixel parameters, the mathematics of computing those partial values is unquestionable. Their rendering engine is handling these **wrong**.)

This fix not only sets a precedent of even more complicated manual shadow definitions, but it also actually fucks up the way these shadows render on Chrome-based browsers. The downgrade isn't huge (as it is balanced to be fairly decent for both), but especially on 4K+ displays like mine, you can tell the faint 1px solid border is there and it doesn't look quite as crisp and "3D" as before this change.

I don't have an issue with having a fix for their bug anyways, but the way this is implemented is quite complicated. It would be nicer if we could code-gen something that doesn't require us to manually add these to each shadow definition.

- Maybe use Sass/SCSS and have a function that creates an optimized pseudo-3D border/shadow given a few parameters?

- Maybe use some sort of browser query to downgrade these shadows to a small selection of basic 1px borders if the user is on Firefox?

---

_Edit:_ Here's an look into what's going on here:

image

Above is a representation of the same box under 3 views.

First, we have a 12 DPR (12 real pixels per logical pixel) view - as you can see, despite the shadow having grown `0.75px` in all directions from the box itself, the `0.5px` offset in the `y` direction has shifted it, such that the top shadow is thinner and the bottom shadow is thicker, relative to the sides. 

This creates the nice pseudo-3D effect, as if this box is a real object being affected by some light source from above, with real physical thickness in its material. The same is true in the correct 1 DPR rendering shown in the middle. These subpixel dark regions are correctly computed into lighter or darker 1-pixel borders. Why the claim that the mathematics is straightforward and not up for debate? Because it is. Each of the four parameters does a very specific thing - perhaps you could have some leeway around the blur parameter, but we aren't using that in these. The other 3 are unquestionable: you duplicate the shape of the box itself, offset it by the X and Y offsets, and expand it in all directions perpendicularly by the fourth thickness parameter. There is only one way to do this.

You can see a recreation of Firefox's incorrect rendering on the right. Somewhere their system is deciding not to draw anything at all above the box, despite there being a 0.25px-thick logical border there. Clearly, then, the result should be a 1px border whose color is 75% that of the background and 25% that of the logical border. But they don't do this at all.

Notably, Firefox understands these subpixel borders. On my 2 DPR display, they correctly render sub(logical)pixel borders down to 0.5px - because there are 4 real pixels in every 1 logical pixel on a 2 DPR display. The issue is they refuse to draw most subpixel values that are below 1 real pixel. This really doesn't make any sense, and there is no good reason to do this.

### DIFF
--- a/gui/script/gen-theme.ts
+++ b/gui/script/gen-theme.ts
@@ -40,7 +40,7 @@ const darkFFRF = (alpha = '15') => ffRenderFix(alpha, 'ffffff');
 const inset = (s: string) => s + ' inset';
 
 // Join with commas.
-const list = (s: string, ...S: string[]) => s + ', ' + S.join(', ');
+const list = (...s: string[]) => s.join(', ');
 
 const themeVariables = {
   'primary-color': ['#000000', '#ffffff', '#ffffff'],


jwmza:
The depressing alternative to any of the above is to do what almost everyone else has done and use a more flat design style with simple 1px solid borders everywhere (e.g. GitHub, VS Code, ...) and totally give up on our nice pseudo-3D border shadows and the subpixel imbalances that give the great feeling we have today.

Personally, I'm a big fan of the way it looks, but there's also nothing wrong with the standard flat rectangles and solid border stuff, so the pain is more about what we'd lose than what we'd end up with.

PR #426, opened 2021-10-19
Rework exo.hcl conversion for docker compose

brandonbloom:
This reworks the docker compose -> exo.hcl conversion to avoid doing any docker-compose magic at the manifest level. Instead, the logic is pushed down in to controllers, so folks who import compose files get nice clean exo.hcl files.


brandonbloom:
This is _almost_ ready. I've reimplemented all the logic _except_ handling the "depends_on" that are implicit in the component spec as service "links" or explicit in "depends_on", but not part of the component-level non-spec stuff.

brandonbloom:
Closing in favor of ongoing work in https://github.com/deref/exo/pull/543

PR #435, opened 2021-10-21
readme: fix 404 on install link

balupton:


brandonbloom:
See https://github.com/deref/exo/commit/1ac8a595863a92c7432e47aafbcb4d5d4c931927 - Thanks!

PR #463, opened 2021-10-29
Make use of the controller-provided dynamic component dependencies

brandonbloom:
This is a follow-up to #436 and is in service of #426.

### DIFF
--- a/internal/core/server/workspace.go
+++ b/internal/core/server/workspace.go
@@ -151,6 +151,23 @@ func (ws *Workspace) Apply(ctx context.Context, input *api.ApplyInput) (*api.App
 	}
 	manifestComponents := componentSet.Components
 
+	manifestComponentDeps := make([][]string, len(manifestComponents))
+	for i, c := range manifestComponents {
+		deps, err := ws.getComponentDeps(ctx, api.ComponentDescription{
+			Name: c.Name,
+			Type: c.Type,
+			Spec: c.Spec,
+		})
+		if err != nil {
+			// If we can't get deps, then just assume that we have none and march
+			// forward. This means that we might fail later when a dependency is
+			// missing, but at least we won't get stuck if there is a provider bug or
+			// some other reason that we can't compute dependencies.
+			ws.logEventf(ctx, "error getting %q deps: %v", c.Name, err)
+		}
+		manifestComponentDeps[i] = deps
+	}
+
 	describeOutput, err := ws.DescribeComponents(ctx, &api.DescribeComponentsInput{})
 	if err != nil {
 		return nil, fmt.Errorf("describing components: %w", err)
@@ -190,7 +207,7 @@ func (ws *Workspace) Apply(ctx context.Context, input *api.ApplyInput) (*api.App
 		allComponents.AddNode(&componentNode{
 			component: c,
 		})
-		for _, dependency := range c.DependsOn {
+		for _, dependency := range manifestComponentDeps[i] {
 			allComponents.AddEdge(c.Name, dependency)
 		}
 	}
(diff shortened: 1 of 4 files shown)


brandonbloom:
This needs a fair amount of testing & it is _very janky_. However, I'm trying to land this janky change, since it unblocks #426, which is badly hurting the UX of the main migration path to exo.hcl files. Assuming this PR does that it's supposed to do, then at least the files on disk won't be crufted up with bad dependency information. Then I'm basically going to totally rework the component system internals in order to fix a variety of bugs. I _think_ I know how to go about that now.

internal/core/server/workspace.go
```diff
@@ -151,6 +151,19 @@ func (ws *Workspace) Apply(ctx context.Context, input *api.ApplyInput) (*api.App
...
 
+	manifestComponentDeps := make([][]string, len(manifestComponents))
+	for i, c := range manifestComponents {
+		deps, err := ws.getComponentDeps(ctx, api.ComponentDescription{
+			Name: c.Name,
+			Type: c.Type,
+			Spec: c.Spec,
+		})
+		if err != nil {
+			ws.logEventf(ctx, "error getting %q deps: %v", c.Name, err)
```
BenElgar, line 162: Why isn't this considered a fatal error?

internal/core/server/workspace.go
```diff
@@ -599,6 +647,7 @@ func (ws *Workspace) RenameComponent(ctx context.Context, input *api.RenameCompo
 
 func (ws *Workspace) RefreshComponents(ctx context.Context, input *api.RefreshComponentsInput) (*api.RefreshComponentsOutput, error) {
 	query := makeComponentQuery(withRefs(input.Refs...))
+	// TODO: Recompute dependencies on refresh.
```
BenElgar, line 650: I'm not quite clear why. Is this because we ignore a failure to get dependencies on apply?

internal/manifest/exohcl/components.go
```diff
@@ -154,29 +153,6 @@ func (c *Component) Analyze(ctx *AnalysisContext) {
 	} else {
 		c.Spec, _ = AnalyzeString(ctx, specAttr.Expr)
 	}
-
-	depsAttr := content.Attributes["depends_on"]
```
BenElgar, line 158: Don't we still need this? How otherwise does a user explicitly define a dependency? How would this docker compose file work?

```docker-compose.yaml
services:
  t0:
    image: bash
    command: "sleep infinity"
  t1:
    image: bash
    command: "sleep infinity"
    depends_on:
      - "t0"
```

internal/core/server/workspace.go
brandonbloom: I'll add a comment. I think that if it were a fatal error, then it would be impossible to stop components that have broken provider implementations or partially invalid specs that fail dependency analysis.

internal/core/server/workspace.go
brandonbloom: I'll expand the comment. The main reason is because the analysis for dependencies in theory may be out of date, either because the provider implementation changed or because it depends on some non-deterministic data (none do currently), so the idea is that "refresh" should fixup and busted state as best it can.

internal/manifest/exohcl/components.go
brandonbloom: depends_on works via https://github.com/deref/exo/pull/436 now - This is simply removing the ability to provide extra/explicit dependencies in the manifest, which I decided isn't really an important feature yet (and if it becomes one, might have to work differently, merging dependencies with the spec analysis, or similar).

brandonbloom:
I did some testing on this and ran in to a problem where dependency resolution can fail in a non-deterministic way because of the janky way in which Apply works now. In short, we delete & recreate things, and then during that interim we might try to resolve names, but the components have been temporarily deleted. This is pretty problematic, and just more evidence that we need to do a major overhaul of the reconciliation engine.

brandonbloom:
This is back to draft, since it's broken. Going to work on overhauling the internals to make this actually easier to get working right.

brandonbloom:
Closing in favor of ongoing work in #543

PR #470, opened 2021-11-03
Add version information to exo api

BenElgar:


internal/core/server/handler.go
```diff
@@ -55,8 +56,16 @@ func BuildRootMux(prefix string, cfg *Config) *http.ServeMux {
 		})
 	}
 
+	versionMiddleware := func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+			w.Header().Add("Exo-Version", cfg.ExoVersion)
```
brandonbloom, line 61: Kinda nitpicky, but isn't this what the `Server` header is for? https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server

BenElgar: I guess — but it's kind of nice not having to parse a version out of a server string. Nbd. Happy to change if you'd prefer.

PR #474, opened 2021-11-03
Add simple example test

BenElgar:


test/tester/tester.go
```diff
@@ -51,21 +53,40 @@ func (et ExoTester) RunTest(ctx context.Context, test ExoTest) (io.Reader, error
...
+		sort.Strings(doneProcesses)
+		if len(names) == len(doneProcesses) && reflect.DeepEqual(names, doneProcesses) {
+			return nil
+		}
+
+		select {
+		case <-ctx.Done():
+			return timeoutError
+		default:
+			time.Sleep(time.Millisecond * 100)
```
brandonbloom, line 87: This should really be something like this:

```go
select {
case <-ctx.Done():
  return timeoutError
case <-time.After(100 * time.Millisecond)
}
```

This way, a timeout will _immediately_ terminate the waiting.

PR #476, opened 2021-11-04
Fix UI upgrade broken when managed

BenElgar:


gui/src/components/VersionInfo.svelte
```diff
@@ -8,22 +8,32 @@
 
   let installedVersion: string | null = null;
   let latestVersion: string | null = null;
+  let isManaged: boolean = false;
   let upgrading = false;
 
   const doUpgrade = async () => {
+    if (isManaged) {
+      alert(
```
brandonbloom, line 16: Do we have nothing better than `alert`?

BenElgar: Not right now. @jwmza

jwmza: Modal system coming imminently, I'm going to wrap up some in-flight things first though.

PR #483, opened 2021-11-05
Add custom template

BenElgar:


BenElgar:
@jwmza Wanted to point out the above where I added a custom Next.js + Prisma + Postgres + Typescript template. It has a tiny messaging app built in.

jwmza:
@BenElgar awesome - excited to see it, I'll have a look ASAP. Power outage in my city block at the moment so I'm unable to do much on my computer temporarily.

jwmza:
Note: we should add the Next.js `IconGlyph` to the data for this template, it will render with a generic icon as-is.

PR #498, opened 2021-11-11
Add simple start counter test

BenElgar:


BenElgar:
This should work in CI (assuming runners aren't reused) but will break after the first run in dev since we don't provide an easy mechanism to clean up volumes.

test/e2etests.go
```diff
@@ -85,6 +86,48 @@ var tests = map[string]tester.ExoTest{
...
+			// FIXME: This doesn't work right now because we don't have an easy way of
+			// removing volumes that are still attached to a previous container.
+			if _, stderr, err := t.RunCmd(ctx, "docker", []string{"volume", "rm", "e2etest-start-counter"}); err != nil {
+				if !strings.Contains(stderr, "No such volume") {
+					return err
+				}
+			}
+
+			if _, _, err := t.RunExo(ctx, "init"); err != nil {
+				return err
```
brandonbloom, line 104: Since this stuff is only used in tests, it's probably perfectly fine to simply panic on error inside `RunExo`.

brandonbloom (approved):

test/e2etests.go
BenElgar: The intent was to allow people to create end to end tests where failure is expected — which becomes a bit ugly if they have to recover from panics all the time.

BenElgar:
This is blocked on https://github.com/deref/exo/issues/499

brandonbloom:
Squashed and cherry-picked into #543

PR #505, opened 2021-11-15
fix syslog for docker on mac

### LINKED ISSUE #478: [BUG] Logs for container components not showing up in GUI or CLI
**Describe the bug**
Container component logs don't show up in the GUI or in the terminal.

**To Reproduce**
1. Create `exo.hcl` with this content:

    ```hcl
    exo = "0.1"

    components {
      container "db" {
        image = "postgres:11.7-alpine"

        environment = {
          POSTGRES_DB = "foo"
          POSTGRES_PASSWORD = "bar"
        }

        ports = ["5432:5432"]
      }
    }
    ```

2. Run `exo init`.
3. Run `exo gui`.

**Expected behavior**
If I use a similar Docker Compose manifest and type `docker compose up`, Docker Compose starts tailing Postgres logs. Based on both the description of the app and the screenshot in the repo README, I expected Exo to do the same. Have I misunderstood how the app works, or did I miss a step?

**System Info (please complete the following information):**
 - OS: macOS 11.6
 - Component: GUI, CLI
 - Version: 2021.10.29

**Additional context**

For `process` components, logs do seem to show up in both the GUI and CLI.

brandonbloom:
Closes #478 

@BenElgar - please make sure everything is still working for you on Linux with this change.

### DIFF
--- a/internal/providers/docker/components/container/lifecycle.go
+++ b/internal/providers/docker/components/container/lifecycle.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os/user"
 	"path"
+	"runtime"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


brandonbloom:
@BenElgar also, it's not clear to me why CI is failing here on the e2e tests. Could you take a peek please?

BenElgar:
It doesn't work:
```
2021/11/16 11:39:59 starting container "3eeeaecdf72239ec98a5a9b80417320f4298beec0f86b9391bd23b0758483234": Error response from daemon: failed to initialize logging driver: dial udp: lookup host.docker.internal: no such host
```

That appears to be why the e2e tests are failing as well.

My understanding is that it is the docker daemon that is resolving the syslog address. On MacOS that daemon runs in a VM where `host.docker.internal` is correctly set but on Linux the daemon is running natively on the host where `host.docker.internal` does not resolve.

The possible solutions I can see are:
1. Detect the operating system and use `localhost` on Linux and `host.docker.internal` on MacOS
2. Examine the network interfaces on the host and look for the one that is named `docker0` and use the attached IP address

brandonbloom:
Gotcha. I'll see if I can figure out some way to detect this more reliably, but can fallback to just checking the OS if there is no obvious API endpoint to deduce this information specifically.

Issue #507, opened 2021-11-15
[UX] submit button below directory picker with long file list is confusing


brandonbloom:
See image:

<img width="1137" alt="image" src="https://user-images.githubusercontent.com/119164/141864999-d217a386-89aa-445a-8502-5f916ce2890e.png">

I didn't realize I had to scroll down at first. Would be nice if the directory list was a scrollable region with a max-height, since I didn't care about any entries in the list, I just wanted to click create where I was at.

jwmza:
Thanks for noting this. I agree - will fix this.

brandonbloom:
Fixed by #523

PR #523, opened 2021-11-17
Height-capped directory browser with scrolling for long lists

jwmza:
This is a working but not ideally styled implementation of scrolling height-capped directory browser, addressing #507 

Could also use this branch to make additional directory browser improvements addressing issues as noted by @brandonbloom earlier today - e.g. symlinks not showing up, and some directories on MacOS not being considered directories by the library.

Another improvement could be to the way the new project directory name input + full path display works. Inline edit should be possible and a UX improvement.

### DIFF
--- a/gui/src/components/DirectoryBrowser.svelte
+++ b/gui/src/components/DirectoryBrowser.svelte
@@ -72,20 +72,15 @@
 
   .directories {
     margin: 12px 0;
+    max-height: 20rem;
+    overflow-y: auto;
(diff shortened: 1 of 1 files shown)


brandonbloom (commented):
> Could also use this branch to make additional directory browser improvements

Can follow-up with that. If you've got small/safe/good changes for main, let's get em merged. Stack PRs if you want to do additional features or reviews, etc.

brandonbloom (approved):

PR #524, opened 2021-11-17
Fix new project root path not being used in CLI-launched GUI

### LINKED ISSUE #408: [BUG] `exo gui` outside of a workspace no longer respects default root path
Go to a non-workspace directory and run `exo gui`, it will take you to the new project page and provide a `root=` query parameter. The GUI no longer makes use of this parameter as a default value when creating a project.

It was even more broken in that it just 404'd, but that is fixed here: https://github.com/deref/exo/commit/d92d1e40e8d5367a4e1a6587b6eb0c66110b24ac

jwmza:
Fixes #408

### DIFF
--- a/gui/src/pages/NewProject.svelte
+++ b/gui/src/pages/NewProject.svelte
@@ -49,7 +49,7 @@
           <button
             on:click={() => {
               router.push(
-                `#/new-project/${template.name}${
+                `#/new-project/${encodeURIComponent(template.name)}${
                   !!root ? `?root=${encodeURIComponent(root)}` : ''
                 }`,
               );
--- a/gui/src/pages/NewProject.svelte
+++ b/gui/src/pages/NewProject.svelte
@@ -9,6 +9,9 @@
   import * as router from 'svelte-spa-router';
   import type { IconGlyph } from '../components/Icon.svelte';
 
+  const queryString = router.querystring;
+  const root = new URLSearchParams($queryString ?? '').get('root');
+
   let search: string = '';
 
   const logoGlyph = (ig: string) => (ig || 'Doc') as IconGlyph;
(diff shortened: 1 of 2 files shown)


jwmza:
Tested & working both when there is and isn't a root parameter given.

Notable: currently this will prompt you to create the New Project as a folder within the folder from which you launched, rather than creating a new project with the folder itself. This might not be ideal but it's a separate issue - if you give no folder name, it should use the working directory as the root of the new project.

gui/src/pages/NewProject.svelte
```diff
@@ -45,7 +48,11 @@
                 .slice(0, search.length) === search.toLocaleLowerCase()) as template}
           <button
             on:click={() => {
-              router.push(`#/new-project/${template.name}`);
+              router.push(
+                `#/new-project/${template.name}${
```
brandonbloom, line 52: Even though we control the template name, we should code defensively here and use an encodeURIComponent on it.

gui/src/pages/NewProjectConfig.svelte
```diff
@@ -28,6 +31,9 @@
   let dirPromise: Promise<ReadDirResult> = api.kernel
     .getUserHomeDir()
     .then((dir) => {
+      if (root) {
+        dir = root;
+      }
       setWorkingDirectory(dir);
       homeDirectory = dir;
       return dirPromise;
```
brandonbloom, line 39: This is really strange... In general, the side effects in this callback seem like they are subverting the Svelte reactive mechanisms.

brandonbloom (approved):

gui/src/pages/NewProjectConfig.svelte
jwmza: Do you man the new 3 lines or the entire mechanism @brandonbloom

brandonbloom: The side effects in this callback in general.

PR #530, opened 2021-11-18
Add readme cmd

BenElgar:


brandonbloom:
I know I mentioned/suggested this idea & thank you for putting it together, but I'm going to close without merging because it seems relatively low priority, has some remaining maintenance/documentation work before anyone can discover/use it, etc. Can reconsider later if we want to improve the devex further.

Issue #324, opened 2021-09-10
[FEATURE] Provide a way to create processes without starting them


ryfow:
**Is your feature request related to a problem? Please describe.**

I have 50ish processes and don't want to start all of them on my laptop right now.

**Describe the solution you'd like**

I'd like a way to define a process without starting it immediately, I don't care strongly about how usage looks.

If I was designing the system, I'd probably completely separate editing the internal manifest from starting things. `apply` and `new process` would alter the internal manifest but you'd need to `start` to get something running.

**Describe alternatives you've considered**

There are lots of options. To preserve backward compatibility, you could add something like a `--define-only` or `--dont-start` options.

You could also change the default behavior but add a `--start` option to do the define & start in one cli invocation.

Brandon said that there are some ideas he's thinking about in #183 but that seems pretty vague at this point. If that's a long ways off, exo could be useful for largish projects soon by splitting defining processes from running them.

brandonbloom:
Thanks for filing this issue. I think the way we're most likely to address this is by adding a run-mode of some kind to the manifest. We're actively working on our own manifest format now, and so will make sure we think about this & will revisit once the new manifest support is available.

ryfow:
It's worth looking at how docker-compose uses profiles. They provide a pretty nice way for users to cluster groups of services. https://docs.docker.com/compose/profiles/

Issue #21, opened 2021-07-21
if component spec is invalid, create still succeeds


brandonbloom:
If you click "Create Process" twice in a row:

![image](https://user-images.githubusercontent.com/119164/126555231-161f8845-e803-4a9b-af8a-b5f99e396f77.png)

![image](https://user-images.githubusercontent.com/119164/126555247-a9e89a71-590b-4e0a-9e00-ef4014e2dd52.png)

Need either/both of earlier spec validation and an "error/invalid" state for components, so that creation will succeed and the component can then be edited after creation. The latter seems more robust.

PR #534, opened 2021-11-22
New internals

### LINKED ISSUE #290: [BUG] apply unnecessarily re-creates resources
Applying a procfile:

```
✓ ├─ deleting hang                                                              
✓ ├─ re-creating hang      
```

But the command and the env didn't change!

We need to be able to handle fully-resolved manifests and compare them against each other to see if any change has happened. If not, then we should avoid recreating.

There is also some complexity here with component dependencies.

brandonbloom:
extremely early, attempting to address or make possible to address numerous issues, including but not limited to:

- #324
- #290
- #21 
- #463 

The general ideas are as follows:

- Factor out "Stack" from "Workspace", specifically the bits in server/workspace.go that might have multiplicity N >= 0 relative to each workspace.
- Refactor "Component" in to physical and logical halves. The logical half will remain called "Component" and the physical half will be called a "Resource". Main distinction is that resources are expected to be external, have a globally unique IRI of some kind, and it is expected that their lifetime is shorter than a component and subject to external influences. Components may have zero or more resources. For example, a process-component will have start & stop, and each time it starts/stops, it will create/destroy a unix-process-resource.
- Keep a pending-operations log for CRUD operations on resources, so that if an operation fails, the user can debug/recover.
- Rework "apply" to use initialize/start/stop/dispose, causing components to create/delete resources, instead of create/delete-ing processes, containers, etc directly. (ie fix #290)
- Write a data migration codepath that takes the component state from above and moves it to some tables in sqlite instead of storing in the state.json file.

### DIFF
--- a/internal/providers/core/resource.go
+++ b/internal/providers/core/resource.go
@@ -1,0 +1,112 @@
+package core
+
+import (
+	"context"
+	"fmt"
+)
+
+type ResourceContext interface {
+	context.Context
+
+	NewController(ctx context.Context, typ string) (ResourceController, error)
+	GetResourceType(ctx context.Context, iri string) (typ string, err error)
+
+	BeginCreate(ctx context.Context, typ string, spec string) (opID string, err error)
+	EndCreate(ctx context.Context, opID string, iri string, err error) error
+
+	BeginUpdate(ctx context.Context, iri string, patch string) (opID string, err error)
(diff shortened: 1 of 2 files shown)


brandonbloom:
going to close this PR for now, as I'm going to come at this problem from a different direction

PR #536, opened 2021-11-24
Add API gateway

BenElgar:


internal/providers/docker/components/apigateway/lifecycle.go
```diff
@@ -0,0 +1,121 @@
...
+
+	token, err := ag.TokenClient.GetToken()
+	if err != nil {
+		return "", fmt.Errorf("getting token: %w", err)
+	}
+
+	return yamlutil.MustMarshalString(container.Spec{
+		Image: compose.MakeString("ghcr.io/deref/exo-mitm:latest"),
+		Environment: compose.Dictionary{Items: []compose.DictionaryItem{
+			{Key: "EXO_URL", Value: "http://host.docker.internal:44643/_exo/"},
```
BenElgar, line 34: Need to figure a way to pass this port in

BenElgar: Done

cmd/mitm/main.go
```diff
@@ -0,0 +1,170 @@
...
+	"net/http/httputil"
+	"net/url"
+	"os"
+	"os/exec"
+	"reflect"
+	"sync"
+	"time"
+
+	"github.com/deref/exo/internal/core/api"
+	"github.com/deref/exo/internal/core/client"
```
BenElgar, line 17: Since this client is internal to the exo package the `mitm` binary needs to be part of the same package. That means that we need to set up a build environment for the entirety of exo in order to be able to build the relevant binary. This makes building the appropriate docker image a great deal slower than it really needs to be. I'd propose moving the client out of the `internal` directory so that we can just build what we need.

.github/workflows/post-commit.yaml
```diff
@@ -0,0 +1,80 @@
...
+            %LocalAppData%\go-build
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+
+      - name: Upgrade npm
+        run: npm i -g npm@latest
+
+      - name: Perform tests
+        run: script/ci-test.sh
```
BenElgar, line 35: This section is just the old `test.yaml` file.

brandonbloom:
Blocked by #543 work.

brandonbloom:
Closing this PR as it is far out of date with the main branch now. Leaving the https://github.com/deref/exo/tree/api-gateway branch around for future reference.

PR #540, opened 2021-12-22
Support boolean docker-compose properties

BenElgar:


internal/manifest/compose/import.go
```diff
@@ -370,6 +370,8 @@ func yamlToHCL(v interface{}) hclsyntax.Expression {
 					return yamlToHCL(v.Value)
 				}
 				return yamlToHCL(i)
+			case "!!bool":
+				return yamlToHCL(v.Value == "true")
```
brandonbloom, line 374: Do we also need to check for `on`, `yes`, etc?

brandonbloom (approved):
LGTM, thanks for the fix!

brandonbloom:
not sure why CI got stuck on this, going to merge & hope it sorts itself out

